PR #18, opened 2017-08-17
Improve error handling when inserting new users into database.

cbailey20043:
The process of inserting new users and retrieving them is now wrapped in
a database transaction. A unique constraint has been placed on the name
field in the activists table to prevent duplicate names.

### DIFF
--- a/model/activist.go
+++ b/model/activist.go
@@ -9,6 +9,20 @@ import (
 	"github.com/jmoiron/sqlx"
 )
 
+/** Constant and Variable Definitions */
+
+const selectUserBaseQuery string = `
+SELECT
+  id,
+  name,
+  email,
+  chapter,
+  phone,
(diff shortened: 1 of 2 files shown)


model/activist.go
```diff
@@ -237,12 +237,47 @@ func GetOrCreateUser(db *sqlx.DB, name string) (User, error) {
...
+
+	_, err = tx.Exec("INSERT INTO activists (name) VALUES (?)", name)
 	if err != nil {
+		tx.Rollback()
 		return User{}, errors.Wrapf(err, "failed to insert user %s", name)
 	}
 
-	return GetUser(db, name)
+	query := `
+SELECT
```
samertm: Can `getUsers` and this use the same "SELECT ... FROM ACTIVISTS" part of the query?

PR #20, opened 2017-08-25
Add "hide activist" feature

samertm:
Add the ability to "suspend" activists, which basically just hides them. They don't show up in the activist list page (eventually we'll add an option to view suspended activists) or in any autocompletes.


main.go
```diff
@@ -129,6 +129,7 @@ func router() *mux.Router {
 	router.Handle("/event/delete", alice.New(main.apiAuthMiddleware).ThenFunc(main.EventDeleteHandler))
 	router.Handle("/activist/list", alice.New(main.apiAuthMiddleware).ThenFunc(main.ActivistListHandler))
 	router.Handle("/activist/save", alice.New(main.apiAuthMiddleware).ThenFunc(main.ActivistSaveHandler))
+	router.Handle("/activist/suspend", alice.New(main.apiAuthMiddleware).ThenFunc(main.ActivistSuspendHandler))
```
mleanos: This is merely for clarification.. 

So the URI pattern of using something like `/activists/:activistId/suspend` isn't being used in this project?

Also in the past I've used a DELETE request method for these types of "soft deletes", and just toggled a field named "deleted" as a boolean or a date-time stamp.

model/activist.go
```diff
@@ -296,3 +327,102 @@ id = :id`, user)
...
+	var userCount int
+	err := db.Get(&userCount, `SELECT count(*) FROM activists WHERE id = ?`, userID)
+	if err != nil {
+		return errors.Wrap(err, "failed to get user count")
+	}
+	if userCount == 0 {
+		return errors.Errorf("User with id %d does not exist", userID)
+	}
+
+	_, err = db.Exec(`UPDATE activists SET suspended = true WHERE id = ?`, userID)
```
mleanos: Related to my other comment, we could just use a field named `deleted` as a Boolean or time-stamp.

It really depends on what the point of "suspend" is. It sounds more like a penalty of some sort based on some business logic, but it's being updated by human intervention.

cbailey20043: "Suspend" does sound like a penalty to a certain degree. In this context, it seems we are hiding an activist from most parts of the UI (the exception being the event attendance page based on my testing). Perhaps `deleted`  or `hidden` would work better, but I'm ok with suspended.

main.go
samertm: We're not using that pattern, but that may be a good idea to adopt later. I think it's built into the mux router that we're using.

Re: using DELETE, our API is an RPC style api, not REST (see https://apihandyman.io/do-you-really-know-why-you-prefer-rest-over-rpc/). There isn't a huge difference between them, I just prefer RPC style because it's more explicit what operations exist, and you don't need to worry about the semantics of different HTTP verbs.

model/activist.go
samertm: I'll change it to `hidden`.

frontend/ActivistList.vue
```diff
@@ -127,30 +161,67 @@ var activistLevelOrder = {
...
+      this.currentModalName = modalName;
+      this.$modal.show(modalName);
     },
     hideModal: function() {
-      this.$modal.hide('edit-activist-modal');
+      if (this.currentModalName) {
+        this.$modal.hide(this.currentModalName);
+      }
+      this.currentModalName = '';
       this.currentActivist = {};
```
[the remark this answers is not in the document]
samertm: Good call

PR #21, opened 2017-08-27
Add "merge activist"

samertm:
Add feature to merge one activist into another.

It works by hiding the activist you're merging and assigning all of their attendance to the "merged" user.

 - Add /activist/merge
 - Add `merged_activist_attendance` table so we can keep track of the merges that have been done so that they can be undone. The undo feature isn't implemented yet, we'll do that once we let people view hidden activists.

Some things I'd like feedback on:
 - Is there a better name than "mergedUserID" for the ID that we're merging into?
 - We don't have any checks that keep people from using the pre-merged user when taking attendance. Should we check that an activist hasn't been merged when we check attendance?

### DIFF
--- a/frontend/ActivistList.vue
+++ b/frontend/ActivistList.vue
@@ -23,6 +23,7 @@
               <button data-role="trigger" class="btn btn-default dropdown-toggle glyphicon glyphicon-option-horizontal" type="button">
               </button>
               <template slot="dropdown">
+                <li><a @click="showModal('merge-activist-modal', activist, index)">Merge Activist</a></li>
                 <li><a @click="showModal('hide-activist-modal', activist, index)">Hide Activist</a></li>
               </template>
             </dropdown>
@@ -38,6 +39,35 @@
       </tbody>
     </table>
     <modal
+       name="merge-activist-modal"
+       :height="600"
+       classes="no-background-color"
+       @opened="modalOpened"
+       @closed="modalClosed"
+       >
+      <div class="modal-dialog">
+        <div class="modal-content">
+          <div class="modal-header">
+            <h2 class="modal-title">Merge activist</h2>
+          </div>
+          <div class="modal-body">
+            <p>Merging activists is used to combine redundant activist entries</p>
+            <p>
+              When you merge this activist into another activist, all of this activist's
+              attendance data will be merged into the other activist. Other data (e.g.
+              email, location, etc) is <strong>NOT</strong> merged.
+            </p>
+            <p>Merge {{currentActivist.name}} into another activist:</p>
+            <select id="merged-activist" class="filter-margin" style="min-width: 200px"></select>
+          </div>
+          <div class="modal-footer">
+            <button type="button" class="btn btn-secondary" @click="hideModal">Close</button>
+            <button type="button" v-bind:disabled="disableConfirmButton" class="btn btn-danger" @click="confirmMergeActivistModal">Merge activist</button>
+          </div>
+        </div>
+      </div>
+    </modal>
+    <modal
        name="hide-activist-modal"
        :height="400"
        classes="no-background-color"
@@ -50,7 +80,7 @@
             <h2 class="modal-title">Hide activist</h2>
           </div>
           <div class="modal-body">
-            <p>Are you sure you want to Hide {{currentActivist.name}}?</p>
+            <p>Are you sure you want to hide {{currentActivist.name}}?</p>
             <p>Hiding an activist hides them from the activist list page but does not delete any event data associated with them. If this activist is a duplicate of another activist, you should merge them instead.</p>
           </div>
           <div class="modal-footer">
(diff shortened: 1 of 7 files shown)


model/activist.go
```diff
@@ -348,6 +348,106 @@ func HideUser(db *sqlx.DB, userID int) error {
...
+	}
+
+	// Let's do this the boring way.
+	var originalUserEventIDs []int
+	err = tx.Select(&originalUserEventIDs, `
+SELECT event_id
+FROM event_attendance
+WHERE
+  activist_id = ?
+`, originalUserID)
```
mleanos: You might be able to simplify the logic that follow this query by changing it to exclude any events that the "merged" (target) user attended.

```sql
SELECT event_id
FROM event_attendance ea
WHERE
  activist_id = ? AND
  NOT EXISTS(
	SELECT ea2.event_id
    FROM event_attendance ea2
    WHERE ea2.activist_id = ? 
		AND ea.event_id = ea2.event_id)
```
Where the second query parameter is the "merged" (target) User.

model/activist.go
```diff
@@ -348,6 +348,106 @@ func HideUser(db *sqlx.DB, userID int) error {
...
+			originalUserID)
+	}
+	for _, eventID := range originalUserEventIDs {
+		// If the merged user also went to the event, just
+		// delete the original user's attendance record.
+		// Otherwise, update the original user's attendance
+		// record so that the activist id is the merged user.
+		var mergedUserAttended int
+		err = tx.Get(&mergedUserAttended, `
+SELECT count(*) FROM event_attendance WHERE activist_id = ? AND event_id = ?
```
mleanos: Related to my comment on the previous query excluding the "merged" User's duplicate events.. 

This wouldn't be needed, if we don't update or delete any of the "source" User's attendance records. By updating or deleting these records, it may complicate things when "undoing" a merge.

Perhaps we could handle this by just adding attendance records for the events (non-duplicate) found in the previous query. The "source" user is already being hidden, and we're tracking the merges. So whenever an undo merge feature is added, it would be just a couple simple operations.

1) Toggle the Activist's "hidden" field.
2) Remove the events that were merged into the "merged" Activist, based on the Activist ID that is getting "re-activated".

I usually try to avoid deletions when there's a possibility that they could be undone in the future.

frontend/event_list.js
```diff
@@ -1,22 +1,5 @@
...
-      /IEMobile/i.test(window.navigator.userAgent) ||
-      /Windows Phone/i.test(window.navigator.userAgent) ||
-      /BlackBerry/i.test(window.navigator.userAgent) ||
-      /BB10/i.test(window.navigator.userAgent) ||
-      /Android.*Mobile/i.test(window.navigator.userAgent)) {
-    return false;
-  }
-  return true;
-}
+import {initActivistSelect} from 'chosen_utils';
```
mleanos: 👍 This is a good practice, and way to adhere to SOC.

model/activist.go
```diff
@@ -348,6 +348,106 @@ func HideUser(db *sqlx.DB, userID int) error {
 	return nil
 }
 
+// Merge originalUserID into mergedUserID.
+//  - The original user is hidden
+//  - All of the original user's event attendance is updated to be the merged user.
+func MergeUser(db *sqlx.DB, originalUserID, mergedUserID int) error {
```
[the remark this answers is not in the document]
samertm: done

model/activist.go
samertm: (One piece of context: hidden activists don't show up on the activist page, but they still get counted for attendance.)

I think our ideas are basically equivalent. You can completely reverse a merge with the info stored in merged_activist_attendance.

The one difference is that for the method you're suggesting, we'd need another field on activists (like `merged`) that indicates that the activist shouldn't show up on the activist page AND shouldn't be counted for attendance. And then when we count attendance, we'll need to join `event_attendance` and `activists` to only include unmerged activists.

model/activist.go
samertm: done

samertm:
thank you for the review!

frontend/ActivistList.vue
```diff
@@ -245,6 +323,18 @@ export default {
       // is shown.
       $(document.body).addClass('noscroll');
       this.disableConfirmButton = false;
+
+      if (this.currentModalName == "merge-activist-modal") {
+        // For some reason, even though this function is supposed to
```
cbailey20043: yeah that's weird.

frontend/chosen_utils.js
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  $.ajax({
+    url: "/activist_names/get",
+    method: "GET",
+    dataType: "json",
+    success: function(data) {
+      var activistNames = data.activist_names;
+
+      activistNames.unshift("");
```
cbailey20043: Why is this needed exactly? Maybe add a brief comment explaining why

samertm: done

model/activist.go
mleanos: So a "hidden" activist still shows up in the list for taking attendance? What's the point of setting the hidden field when merging then?

PR #23, opened 2017-09-06
Add prompt for the developers email to `make dev_db`

cjrd:
When running `make dev_db`, the user is now prompted for their development email, and it is injected into `adb_db.adb_users` as an admin. Alternatively, the user can set the environment variable `DXE_DEV_EMAIL` in e.g. their `.bashrc` and never see the prompt.

I left the existing emails in the database creation script, though I can pull these out if you'd like to make this script a little cleaner =)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,7 +23,7 @@ watch:
 # Wipe and re-create the dev databases. See the readme for more
 # details.
 dev_db:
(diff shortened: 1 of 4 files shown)


samertm:
This is great!

samertm:
cc @mleanos

PR #24, opened 2017-09-15
feat(admin): User management

mleanos:
Adds an Admin feature for managing ADB Users.

~~This is still a WIP, but has most of the functionality.~~ Modeled
implementation after Activists feature.

Currently lists Users, and allows adding/updating a User,

TODO:
Allow Deleting a User
Update Admin & Disabled from quick menu button (like Hide & Merge for
Activists)
Fix spacing

### DIFF
--- a/frontend/UserList.vue
+++ b/frontend/UserList.vue
@@ -172,55 +172,6 @@ export default {
         limit: 40
       }
     },
-    /*
-    removeUser: function () {
-      // Disable the save button
-      this.disableConfirmButton = true;
-      
-      if (!window.confirm("Are you sure you want to delete this user?")) {
-        this.disableConfirmButton = false;
-        return;
-      }
-      
-      $.ajax({
-        url: "/user/delete",
-        method: "POST",
-        contentType: "application/json",
-        data: JSON.stringify(this.currentUser),
-        success: (response) => {
-          // Handle successful response
-          
-          var parsed = JSON.parse(response);
-          
-          if (parsed.status !== "success") {
-            flashMessage("Error: " + parsed.message, true);
-            return;
-          }
-          
-          flashMessage(this.currentUser.email + " removed");
-          
-          // Remove user from list
-          var users = this.users
-            .slice(0, this.userIndex)
-            .concat(
-              this.users
-                .slice(this.userIndex + 1)
-             );
-             
-          this.setUsers(users);
-             
-          this.disableConfirmButton = false;
-          this.hideModal();
-        },
-        error: (errorResponse) => {
-          this.disableConfirmButton = false;
-          
-          console.warn(errorResponse.responseText);
-          flashMessage("Server error: " + err.responseText, true);
-        }
-      });
-    }
-    */
   },
   data() {
     return {
--- a/frontend/UserList.vue
+++ b/frontend/UserList.vue
@@ -1,0 +1,250 @@
+<template>
+
+  <div id="app" class="main">
+    <button class="btn btn-default" @click="showModal('edit-user-modal')">Add New User</button>
+    <table id="user-list" class="adb-table table table-hover table-striped tablesorter">
+      <thead>
+        <tr>
+          <th></th>
+          <th></th>
+          <th>Email</th>
+          <th>Admin</th>
+          <th>Disabled</th>
+        </tr>
+      </thead>
+      <tbody id="user-list-body">
+        <tr v-for="(user, index) in users">
+          <td><button class="btn btn-default glyphicon glyphicon-pencil" @click="showModal('edit-user-modal', user, index)"></button></td>
+          <td>
+            <dropdown>
+              <button data-role="trigger" class="btn btn-default dropdown-toggle glyphicon glyphicon-option-horizontal" type="button">
+              </button>
+            </dropdown>
+          </td>
+          <td>{{user.email}}</td>
+          <td>{{user.admin}}</td>
+          <td>{{user.disabled}}</td>
+        </tr>
+      </tbody>
+    </table>
+    <modal
+       name="edit-user-modal"
+       :height="830"
+       classes="no-background-color"
+       @opened="modalOpened"
+       @closed="modalClosed"
+       >
+      <div class="modal-dialog">
+        <div class="modal-content">
+          <div class="modal-header">
+            <h2 class="modal-title" v-if="currentUser.id">Edit user</h2>
+            <h2 class="modal-title" v-if="!currentUser.id">New user</h2>
+          </div>
+          <div class="modal-body">
+            <form action="" id="editUserForm">
+              <p><label for="email">Email: </label><input class="form-control" type="text" v-model.trim="currentUser.email" id="email" /></p>
+              <p><label for="admin">Admin: </label><input class="form-control" type="checkbox" v-model="currentUser.admin" id="admin" /></p>
+              <p><label for="disabled">Disabled: </label><input class="form-control" type="checkbox" v-model="currentUser.disabled" id="disabled" /></p>
+            </form>
+          </div>
+          <div class="modal-footer">
+            <button type="button" class="btn btn-secondary" @click="hideModal">Close</button>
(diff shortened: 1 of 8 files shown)


mleanos:
@samertm @cbailey20043 Care to review what I have thus far?

samertm:
Looks great so far!

Re: spacing, you can use gofmt to automatically format you go code: https://blog.golang.org/go-fmt-your-code

mleanos:
Here are some screen shots of this new feature..

image

image

image


Should this dropdown button provide a quick way to toggle the "Admin" & "Disabled" flags?

image

frontend/UserList.vue
```diff
@@ -0,0 +1,298 @@
...
+    showModal: function (modalName, user, index) {
+      // Check to see if there's a modal open, and close it if so.
+      if (this.currentModalName) {
+        this.hideModal();
+      }
+
+      // Make shallow copy of selected activist to prevent persisting unsaved
+      // edits at the view layer when closing modal
+      if (user) {
+        this.currentUser = $.extend({}, user);
```
samertm: `this.currentUser = $.extend({}, user);` works even if `user` doesn't exist

frontend/UserList.vue
```diff
@@ -0,0 +1,298 @@
...
+        success: (data) => {
+          this.disableConfirmButton = false;
+
+          var parsed = JSON.parse(data);
+          if (parsed.status === "error") {
+            flashMessage("Error: " + parsed.message, true);
+            return;
+          }
+          // status === "success"
+          console.log('saved: ', parsed);
```
samertm: Remove these logs?

main.go
```diff
@@ -180,6 +188,34 @@ func (c MainController) authMiddleware(h http.Handler) http.Handler {
...
+				HttpOnly: true,
+			}
+			http.SetCookie(w, c)
+
+			http.Redirect(w, r, "/login", http.StatusFound)
+			return
+		}
+
+		if !user.Admin {
+			// Add 403 status
```
samertm: Maybe it's good enough to flash that the user doesn't have permission to access that route.

main.go
```diff
@@ -576,6 +634,78 @@ func (c MainController) PowerHandler(w http.ResponseWriter, r *http.Request) {
 	})
 }
 
+func (c MainController) UserListHandler(w http.ResponseWriter, r *http.Request) {
+	users, err := model.GetUsersJSON(c.db, model.GetUserOptions{})
+
+	if err != nil {
+		panic(err)
```
samertm: do this instead?

```
sendErrorMessage(w, err)
return
```

model/adb_auth.go
```diff
@@ -44,3 +72,191 @@ FROM adb_users
...
+
+func CleanUserData(body io.Reader) (ADBUser, error) {
+	var userJSON UserJSON
+	err := json.NewDecoder(body).Decode(&userJSON)
+
+	if err != nil {
+		return ADBUser{}, err
+	}
+
+	if err := checkForDangerousChars(userJSON.Email); err != nil {
```
samertm: We probably don't need this check. We only need it for data that ends up on non-vue pages, like new event and events list, because those pages don't properly escape values from the server.

frontend/UserList.vue
```diff
@@ -0,0 +1,298 @@
...
+      this.userIndex = -1;
+      this.disableConfirmButton = false;
+      this.currentModalName = '';
+      this.pagingParameters = {
+        email: "",
+        order: AscOrder,
+        limit: 40
+      }
+    },
+    removeUser: function () {
```
samertm: Now that I think about it, I don't think we need to expose deleting users through the UI. If someone wants to delete an account, they should just disable that user. WDYT?

templates/header.html
```diff
@@ -39,7 +39,9 @@
             <li class="{{if (eq .PageName "ActivistList")}}active{{end}}"><a href="/list_activists">Activists</a></li>
             <li class="{{if (eq .PageName "Leaderboard")}}active{{end}}"><a href="/leaderboard">Leaderboard</a></li>
             <li class="{{if (eq .PageName "Power")}}active{{end}}"><a href="/power">Power</a></li>
+	    <li class="{{if (eq .PageName "UserList")}}active{{end}}"><a href="/admin/users">Users</a></li>
```
samertm: Is it possible to only show this to admins? ... probably not unless each request has access to the user, which would be a pain in the butt to add unless we have request-local contexts: https://joeshaw.org/revisiting-context-and-http-handler-for-go-17/

samertm:
LGTM! Left a couple comments

frontend/UserList.vue
mleanos: Sure thing. Forgot about those.

model/adb_auth.go
mleanos: I didn't know that about Vue escaping input values. Good to know :) 

I was surprised to see the need for this, as I thought the router would have something built in. I'll go ahead and remove it. Especially since only "admins" will be using the endpoint that calls this method.

frontend/UserList.vue
mleanos: I was thinking about it this way..

The `disable` flag would most likely be used as a temporary measure. Whereas a "deletion" would probably be something is permanent, and it seems like a more common use case. We could always do a "soft delete", where we set a "deleted" flag.

main.go
mleanos: We probably don't want the user to be sitting on that page, or "be allowed to end up there", when they don't have permission. Ideally, we might want to send them to a new page that shows conveys that they don't have access to that page. Something like a "403" route.

The `apiAdminAuthMiddleware` actually only flashes that they don't have permission.

frontend/UserList.vue
mleanos: Good catch. I'll change that.

templates/header.html
mleanos: Thanks for the link. I'll look at the context stuff.

frontend/UserList.vue
samertm: Ah, I had `disabled` in mind as the permanent feature. If we just want to delete folks, which is fine, you should remove the `disabled` flag.

main.go
samertm: Ah, I see. Want to create a 403 route?

frontend/UserList.vue
mleanos: Ok. I guess I hadn't understood the purpose of `disabled`. I thought of it as a way to temporarily disable an account (for whatever reasons); I could see some scenarios where that might be useful.

If we just need a way to remove the user, and not what I described above, then just using the `disabled` would work fine. In that case, I can remove the "delete" button from the UI & disable the api endpoint. (while leaving the http handler & model methods for possible future use).

main.go
mleanos: Yea. I can do that.

frontend/UserList.vue
```diff
@@ -0,0 +1,284 @@
...
+      this.disableConfirmButton = false;
+      this.currentModalName = '';
+      this.pagingParameters = {
+        email: "",
+        order: AscOrder,
+        limit: 40
+      }
+    },
+    /*
+    removeUser: function () {
```
samertm: Delete this code if we don't need it?

main.go
```diff
@@ -111,6 +111,7 @@ func router() *mux.Router {
 	router := mux.NewRouter()
 	// Unauthed pages
 	router.HandleFunc("/login", main.LoginHandler)
+  router.HandleFunc("/403", main.ForbiddenHandler)
```
samertm: gofmt?

mleanos: Yep.. Was just waiting for the approval of this PR, to keep the diffs easy to notice without all the formatting changes.

mleanos:
@samertm Thanks :) I'll format the code, and let you merge when you're ready.

Probably should use the squash merge option here. Not sure if that's standard practice with this project though.

mleanos:
@samertm Made a couple other minor changes for code consistency, and one change to the `/403` route to add the User Context.

PR #25, opened 2017-09-28
Add working groups page

samertm:
Adds the working group page.

 - Adds tables `working_groups` and `working_group_members`.
 - Vendors a fork of https://github.com/moreta/vue-search-select, modified to be similar to chosen-js.

Will follow up with a pull request that adds mailing list syncing.

### DIFF
--- a/model/db.go
+++ b/model/db.go
@@ -85,6 +85,7 @@ CREATE TABLE IF NOT EXISTS working_groups (
   name VARCHAR(60) NOT NULL,
   type TINYINT(1) NOT NULL,
   group_email VARCHAR(100),
+  point_person_id INTEGER NOT NULL DEFAULT '0',
   CONSTRAINT working_groups_name_ukey UNIQUE (name)
 )
 `)
@@ -93,7 +94,6 @@ db.MustExec(`
 CREATE TABLE IF NOT EXISTS working_group_members (
   working_group_id INTEGER NOT NULL,
   activist_id INTEGER NOT NULL,
-  point_person TINYINT(1) NOT NULL DEFAULT '0',
   CONSTRAINT working_group_member_ukey UNIQUE (working_group_id, activist_id)
 )
 `)
--- a/model/db.go
+++ b/model/db.go
@@ -23,6 +23,8 @@ func WipeDatabase(db *sqlx.DB) {
 	db.MustExec(`DROP TABLE IF EXISTS event_attendance`)
 	db.MustExec(`DROP TABLE IF EXISTS adb_users`)
 	db.MustExec(`DROP TABLE IF EXISTS merged_activist_attendance`)
+  db.MustExec(`DROP TABLE IF EXISTS working_groups`)
+  db.MustExec(`DROP TABLE IF EXISTS working_group_members`)
 
 	db.MustExec(`
 CREATE TABLE IF NOT EXISTS activists (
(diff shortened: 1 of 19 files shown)


model/db.go
```diff
@@ -76,6 +78,32 @@ CREATE TABLE IF NOT EXISTS merged_activist_attendance (
...
+`)
+
+	db.MustExec(`
+CREATE TABLE IF NOT EXISTS working_group_members (
+  working_group_id INTEGER NOT NULL,
+  activist_id INTEGER NOT NULL,
+  -- True if the activist is the point person of the working group.
+  -- There should be only one point person per working group, but we
+  -- don't restrict that on the backend.
+  point_person TINYINT NOT NULL DEFAULT '0',
```
cbailey20043: What was the reason for moving the point person field back to the working_group_members? I felt it was much easier to enforce the "single point person per group" constraint this way.

samertm:
I felt like it made the code easier to reason about b/c it makes
accessing + modifying working group members more uniform. Also, when
point_person is part of the working groups it's less clear what the
relationship is between the point person and the rest of
working_group_members (should the point person always exist in
working_group_members, or should it never?)

On Wed, Sep 27, 2017, at 09:08 PM, cbailey20043 wrote:
> *@cbailey20043* commented on this pull request.


> 
> In model/db.go[1]:


> > +  id INTEGER PRIMARY KEY AUTO_INCREMENT, +  name VARCHAR(60) NOT
> >    NULL, +  type TINYINT(1) NOT NULL, +  group_email VARCHAR(100)
> >    NOT NULL, +  CONSTRAINT working_groups_name_ukey UNIQUE (name) +)
> >    +`) + + db.MustExec(` +CREATE TABLE IF NOT EXISTS
> >    working_group_members ( +  working_group_id INTEGER NOT NULL, +
> >    activist_id INTEGER NOT NULL, +  -- True if the activist is the
> >    point person of the working group. +  -- There should be only one
> >    point person per working group, but we +  -- don't restrict that
> >    on the backend. +  point_person TINYINT NOT NULL DEFAULT '0',
>> What was the reason for moving the point person field back to the
> working_group_members? I felt it was much easier to enforce the
> "single point person per group" constraint this way.> — You are receiving this because you authored the thread. Reply to
> this email directly, view it on GitHub[2], or mute the thread[3].> 

Links:

  1. https://github.com/dxe/adb/pull/25#discussion_r141522525
  2. https://github.com/dxe/adb/pull/25#pullrequestreview-65747721

model/working_groups.go
```diff
@@ -0,0 +1,431 @@
...
+	}
+	return workingGroup.ID, nil
+}
+
+func insertWorkingGroupMembers(tx *sqlx.Tx, workingGroup WorkingGroup) error {
+	if workingGroup.ID == 0 {
+		return errors.New("Invalid WorkingGroup ID. ID's must be greater than 0")
+	}
+	// First drop all working group members for the working group.
+	_, err := tx.Exec(`DELETE FROM working_group_members WHERE working_group_id = ?`, workingGroup.ID)
```
cbailey20043: Good idea. Completely forgot about doing that myself

samertm:
> The one general comment I have (and you should go ahead and ship regardless if you agree or not, since its not a major thing) is that it seems a tad confusing that the point person is not listed as a member.

Ah, that's a good point. I'll ship this and we'll keep that in mind to fix in the future, since I'm not sure how that would work right now.

> Finally, I would probably let Michael review as well before shipping, since there are a lot of changes, and he may catch something that I missed (especially in the front end).

That's probably a good idea, but I'm eager to merge this :P Michael, if you have any comments feel free to leave them and I'll address them in another commit.

PR #32, opened 2018-12-24
Rewrite events/connections list and create/edit pages to use Vue

mdempsky:
First major step towards making ADB into a single-page Vue app: migrate the remaining server-side templates into Vue pages.


mdempsky:
I've tested this quite a bit locally, and believe everything works correctly. It would be great if someone could double check that.

jakehobbs:
I found three issues:

Can we start the attendee list as having 5 rows instead of 1?

After I save a new event, the top attendee field is yellow:

screen shot 2018-12-26 at 1 26 01 pm

Show full attendee list by default on list_connections.

mdempsky:
> The ActivistList selector on the Event & Maintenance list views aren't showing the selected Activist in the dropdown.

Hm, I'm not able to reproduce this issue. The Activist filter option seems to work how I'd expect. Can you take a screenshot?

> Getting console errors complaining about a missing prop in the Vue in some of the Circles, Chapter Members, and Organizers list views. (see below screenshot)

This is because [the validViews validator in ActivistList.vue](https://github.com/dxe/adb/blob/8addaaeec217c1c2eb9215415f420c057fdc189b/frontend/ActivistList.vue#L820) is out of date. Unrelated to this PR.

> One thing that's jumping out at me is the reuse of the event-list directive for Connections. 

As discussed on WhatsApp, this is awkward, but it's how things are currently done already. Hopefully moving to Vue makes it a bit easier to fix this though.

> Can we start the attendee list as having 5 rows instead of 1?

Done. I guess I didn't see the point in having extra blank rows, but I've changed the logic to now be there are always at least 5 rows, and a blank row at the bottom.

> After I save a new event, the top attendee field is yellow:

I'm guessing what happened was you edited an event to replace the top name with a new activist name? If so, this should be fixed now. I wasn't updating the set of known-activists after saving the event, nor was I refreshing the UI.

> Show full attendee list by default on list_connections.

Done.

PR #34, opened 2018-12-24
Update npm dependencies

### LINKED ISSUE #33: npm dependencies out of date
It looks like a lot of our npm dependencies are out of date. For example, webpack is up to version 4 now, whereas we're still using version 2. handsontable is at version 6.2, but we're still using 0.34.

I'm also a little confused about the frontend/external directory. I'd think we should just be accessing awesomplete/handsontable-vue/etc from node_modules.

mdempsky:
Two notable changes:

1. webpack now natively support development and production build "modes",
which obviates the need for manually configuring additional minimization
plugins in production mode. So our webpack.config.js file simplifies a bit.
(Exception: vue-loader now requires a webpack plugin.)

2. babel config now automatically detects which modern JavaScript constructs
need to be transpiled to support legacy browsers.

While here, remove unused vue-infinite-loading dependency.

Fixes #33.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,6 +1,5 @@
 {
     "presets": [
-        ["stage-2"],
-        ["es2015",  {"modules": false}]
+        ["@babel/preset-env"]
     ]
 }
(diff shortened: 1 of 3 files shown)


jakehobbs:
Is there a way to reopen this?

PR #39, opened 2018-12-28
Feature/users roles

mleanos:
Implements a roles based User access feature. Currently, any user has access to all areas of this application; with the exception of "admins" who have access to the Users management feature. These changes will provide a way to restrict access to certain areas of the application by way of a new table named `users_roles`.

Initially we can start with only providing roles for elevated access. If there is no row in `users_roles` for a specific user, we can treat them as a user with minimal access; i.e. they can only access the Attendance taking feature where they can create an Event and add Attendees.

This is still a work in progress..

TODO's:

-  Backend
    1. ~~User Roles Update~~
    2. ~~User Roles Remove~~
    3. ~~Retrieve roles for authenticated user~~

- Frontend
    1. ~~Add Roles management to users management (admin)~~
    2. ~~Restrict access based on currently authenticated user's roles.~~

Satisfies: #38

PR #41, opened 2019-01-06
feat(users): User Role Management

mleanos:
Implements a User Role based authentication management, and Admin UI to assign roles.

These changes were originally introduced into master, but had to be reverted. For a full history of the changes and discussion see the below PR.

https://github.com/dxe/adb/pull/39

### DIFF
--- a/frontend/UserList.vue
+++ b/frontend/UserList.vue
@@ -7,16 +7,16 @@
         <tr>
           <th></th>
           <th>Email</th>
-          <th>Admin</th>
           <th>Disabled</th>
+          <th>Roles</th>
         </tr>
       </thead>
       <tbody id="user-list-body">
         <tr v-for="(user, index) in users">
           <td><button class="btn btn-default glyphicon glyphicon-pencil" @click="showModal('edit-user-modal', user, index)"></button></td>
           <td>{{user.email}}</td>
-          <td>{{user.admin}}</td>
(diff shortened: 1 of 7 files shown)


mleanos:
@jakehobbs @MarkMcCaskey  Should we try this again?

mleanos:
@jakehobbs Regarding your questions as to why store the roles in a separate table, and not just use columns in the adb_user table..

From a data model perspective it makes more sense for the Roles to be a list (array) because we don't know how many roles we'll end up having in the future and how many roles a User would theoretically be assigned to. If we used columns then we could end up building logic (front-end or back-end) that can paint us into a corner; making it harder to extend our Roles based authentication.

I know that creating this new table, and the model methods & implementation seems like a lot of overhead at this point, but it feels like the best practice.

jakehobbs:
> @jakehobbs Regarding your questions as to why store the roles in a separate table, and not just use columns in the adb_user table..
> 
> From a data model perspective it makes more sense for the Roles to be a list (array) because we don't know how many roles we'll end up having in the future and how many roles a User would theoretically be assigned to. If we used columns then we could end up building logic (front-end or back-end) that can paint us into a corner; making it harder to extend our Roles based authentication.
> 
> I know that creating this new table, and the model methods & implementation seems like a lot of overhead at this point, but it feels like the best practice.

Sounds good!

PR #42, opened 2019-01-06
Use Awesomplete from npm

mdempsky:


mdempsky:
I tracked down through commit history to determine that frontend/external/awesomplete.js was https://github.com/LeaVerou/awesomplete/blame/7b17750802ee3a73d9c19f5dca1da71648f52a36/awesomplete.js with two changes:

1. We added support to treat "tab" as "enter".

2. We added custom filtering logic.

The second is easy enough to fix: just set the `filter` option to our own logic (which I simplified too).

The first is slightly trickier. Upstream has subsequently added support for select-on-TAB, but we had patched it to cancel the event's default behavior (i.e., don't advance focus automatically), whereas upstream left it alone. This conflicts a little with our custom logic to advance focus, as it can result in focus being advanced twice.

The simplest fix I found was to wait until next tick to advance focus, so that the tab event has already settled.

PR #43, opened 2019-01-07
faet(auth): User logout

mleanos:
Adds logout functionality for users to logout of the currently signed in Google account.

We clear the server-side Auth Cookie, as well as log them out of the Google Sign In.


mdempsky:
Looked over the code and it looks good.

PR #45, opened 2019-01-09
Convert Vue components to class-style

mdempsky:
The diff looks large, but it's mostly mechanical: instead of

    export default {
      data() {
        return {
          x: 42,
          y: "quux",
        };
      },
      created() { ... },
      methods: {
        foo() { ... },
        bar() { ... },
      },
    };

we can now write components as (almost) normal TypeScript classes:

    @Component
    export default class MyComponent extends Vue {
      x = 42;
      y = "quux";

      created() { ... }
      foo() { ... }
      bar() { ... }
    }

There are a couple minor gotchas to be aware of, but they're reasonably documented:

https://vuejs.org/v2/guide/typescript.html#Class-Style-Vue-Components
https://github.com/vuejs/vue-class-component

### DIFF
--- a/frontend/ActivistList.vue
+++ b/frontend/ActivistList.vue
@@ -206,7 +206,7 @@
 import vmodal from 'vue-js-modal';
 import HotTable from './external/vue-handsontable-official/HotTable.vue';
 import { rewriteSettings } from './external/vue-handsontable-official/helpers';
-import Vue from 'vue';
+import { Vue, Component, Prop, Watch } from 'vue-property-decorator';
 import { focus } from './directives/focus';
 import { flashMessage } from './flash_message';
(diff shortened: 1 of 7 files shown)


mdempsky:
If you want to review the changes, I suggest using `git diff`'s `--ignore-space-change` flag.

Otherwise, just take a look at how the code looks now, and let me know if you think the class style is an improvement over what we had before.

jakehobbs:
Sorry, somehow I missed this. I'm fine with it if the conflicts are resolved.

jakehobbs:
@mdempsky do we want to revisit this?

mdempsky:
Probably, but I'm going to close for now. I'll reopen once I have time to look at it again. It's not urgent.

PR #47, opened 2019-02-08
Contact info indicators

jakehobbs:


jakehobbs:
Performance is too slow. Rolling back.

PR #57, opened 2019-12-01
remove unused CSS class names

mdempsky:
For each removed CSS class name "foo", I couldn't find ".foo" anywhere
in a CSS source file.


jakehobbs:
LGTM

PR #59, opened 2019-12-12
integrate surveys into adb

jakehobbs:
Currently, we have mysql scripts that run on a regular basis to export event attendance data to csv files. These csv files are publicly accessible and imported into Google Sheets. From there, there are multiple Google Apps Scripts that run to send survey emails via Gmail.

This pull request will add all of the survey-sending functionality directly into the ADB.

Here are the steps I will follow when merging this:
- Disable current GAS scripts on various spreadsheets: meetup, chapter mtg, protest.
- Add survey_sent col to prod events database (and set to 1 for previous day's events to prevent sending duplicates).
- Add new environment variables to docker instance (see readme).
- Remove surveys.sql from mysql-files and delete the csv files from www.

### DIFF
--- a/model/event.go
+++ b/model/event.go
@@ -46,14 +46,16 @@ type EventJSON struct {
 
 /* TODO Restructure this Struct */
 type Event struct {
-	ID               int       `db:"id"`
-	EventName        string    `db:"name"`
-	EventDate        time.Time `db:"date"`
-	EventType        EventType `db:"event_type"`
-	Attendees        []string  // For retrieving all event attendees
-	AttendeeEmails   []string
-	AddedAttendees   []Activist // Used for Updating Events
-	DeletedAttendees []Activist // Used for Updating Events
+	ID                    int       `db:"id"`
+	EventName             string    `db:"name"`
+	EventDate             time.Time `db:"date"`
+	EventType             EventType `db:"event_type"`
+	SurveySent            int       `db:"survey_sent"` // Used for sending event surveys
+	Attendees             []string  // For retrieving all event attendees
+	AttendeeEmails        []string
+	AttendeeMissingEmails []string   // Used for sending event surveys
+	AddedAttendees        []Activist // Used for Updating Events
+	DeletedAttendees      []Activist // Used for Updating Events
 }
 
 func (event *Event) ToJSON() EventJSON {
@@ -77,6 +79,7 @@ type GetEventOptions struct {
 	EventType      string
 	EventNameQuery string
 	EventActivist  string
+	SurveySent     string
 }
 
 /** Functions and Methods */
@@ -116,7 +119,7 @@ func GetEvent(db *sqlx.DB, options GetEventOptions) (Event, error) {
 
 func getEvents(db *sqlx.DB, options GetEventOptions) ([]Event, error) {
 	var queryArgs []interface{}
-	query := `SELECT e.id, e.name, e.date, e.event_type FROM events e `
+	query := `SELECT e.id, e.name, e.date, e.event_type, e.survey_sent FROM events e `
 
 	// Items in whereClause are added to the query in order, separated by ' AND '.
 	var whereClause []string
@@ -145,6 +148,10 @@ ON (e.id = ea.event_id AND ea.activist_id = a.id)
 		whereClause = append(whereClause, "e.date <= ?")
 		queryArgs = append(queryArgs, options.DateTo)
 	}
+	if options.SurveySent != "" {
+		whereClause = append(whereClause, "e.survey_sent = ?")
+		queryArgs = append(queryArgs, options.SurveySent)
+	}
 	if options.EventType == "noConnections" {
 		whereClause = append(whereClause, "e.event_type <> 'Connection'")
 	} else if options.EventType == "mpiDA" {
@@ -152,7 +159,7 @@ ON (e.id = ea.event_id AND ea.activist_id = a.id)
 	} else if options.EventType == "mpiCOM" {
 		whereClause = append(whereClause, "(e.event_type = 'Community' or e.event_type = 'Training' or e.event_type = 'Circle')")
 	} else if options.EventType != "" {
-		whereClause = append(whereClause, "e.event_type = ?")
+		whereClause = append(whereClause, "e.event_type like ?")
 		queryArgs = append(queryArgs, options.EventType)
 	}
 	if options.EventNameQuery != "" {
@@ -219,7 +226,11 @@ WHERE
 	for _, a := range allAttendance {
 		i := eventIDToIndex[a.EventID]
 		events[i].Attendees = append(events[i].Attendees, a.ActivistName)
-		events[i].AttendeeEmails = append(events[i].AttendeeEmails, a.ActivistEmail)
+		if a.ActivistEmail != "" {
+			events[i].AttendeeEmails = append(events[i].AttendeeEmails, a.ActivistEmail)
+		} else {
+			events[i].AttendeeMissingEmails = append(events[i].AttendeeMissingEmails, a.ActivistName)
+		}
 	}
 
 	return events, nil
@@ -355,6 +366,41 @@ WHERE
 		tx.Rollback()
 		return 0, errors.Wrap(err, "failed to insert event attendance")
 	}
+	if err := tx.Commit(); err != nil {
+		tx.Rollback()
+		return 0, errors.Wrap(err, "failed to commit update event")
+	}
+	return event.ID, nil
+}
+
+func UpdateEventSurveyStatus(db *sqlx.DB, event Event) (eventID int, err error) {
+	tx, err := db.Beginx()
+	if err != nil {
+		return 0, errors.Wrap(err, "failed to update event")
+	}
+	// Error out if the event doesn't exist.
+	var eventCount int
+	err = tx.Get(&eventCount, `SELECT count(*) FROM events WHERE id = ?`, event.ID)
+	if err != nil {
+		tx.Rollback()
+		return 0, errors.Wrap(err, "failed to get event count")
+	}
+	if eventCount == 0 {
+		tx.Rollback()
+		return 0, errors.Errorf("Event with id %d does not exist", event.ID)
+	}
+
+	// Update the event
+	_, err = tx.NamedExec(`UPDATE events
+SET
+  survey_sent = :survey_sent
+WHERE
+  id = :id`, event)
+	if err != nil {
+		tx.Rollback()
+		return 0, errors.Wrap(err, "failed to update event")
+	}
+
 	if err := tx.Commit(); err != nil {
 		tx.Rollback()
 		return 0, errors.Wrap(err, "failed to commit update event")
--- a/model/event.go
+++ b/model/event.go
@@ -378,17 +378,6 @@ func UpdateEventSurveyStatus(db *sqlx.DB, event Event) (eventID int, err error) 
 	if err != nil {
 		return 0, errors.Wrap(err, "failed to update event")
 	}
-	// Error out if the event doesn't exist.
-	var eventCount int
-	err = tx.Get(&eventCount, `SELECT count(*) FROM events WHERE id = ?`, event.ID)
-	if err != nil {
-		tx.Rollback()
-		return 0, errors.Wrap(err, "failed to get event count")
-	}
-	if eventCount == 0 {
-		tx.Rollback()
-		return 0, errors.Errorf("Event with id %d does not exist", event.ID)
-	}
 
 	// Update the event
 	_, err = tx.NamedExec(`UPDATE events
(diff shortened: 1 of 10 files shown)


survey_mailer/survey_mailer.go
```diff
@@ -0,0 +1,204 @@
...
+
+func surveyMailerWrapper(db *sqlx.DB) {
+	defer func() {
+		if r := recover(); r != nil {
+			log.Println("Recovered from panic in survey mailer", r)
+		}
+	}()
+
+	// Get current hour of day & current day of week
+	weekday := time.Now().Weekday()
```
mdempsky: nit: I'd do:

```
now := time.Now()
```

and change all of these to reuse that same now value. Otherwise, you can have weird race conditions if this function happens to get called at a time boundary.

survey_mailer/survey_mailer.go
```diff
@@ -0,0 +1,204 @@
...
+
+// Sends surveys based on event attendance every 60 minutes.
+// Should be run in a goroutine.
+func StartSurveyMailer(db *sqlx.DB) {
+	for {
+		log.Println("Starting survey mailer")
+		surveyMailerWrapper(db)
+		log.Println("Finished survey mailer")
+		// TODO(jake): change this to 60 min before deploying
+		time.Sleep(60 * time.Minute)
```
mdempsky: I'd write this as:

    for now := range time.Tick(time.Hour) {
        ...
        surveyMailerWrapper(db, now)
        ...
    }

That ensures the clock doesn't drift over time.

I'm somewhat concerned there's still a chance that you get one event at 8:59:59.999999 and then because of timer delays, the next event doesn't happen until 10:00:00.000000 and you completely miss the events at 9am. I'll dig into the timer code to see if this is a risk.

survey_mailer/survey_mailer.go
```diff
@@ -0,0 +1,204 @@
...
+	// send meetup surveys on saturdays at 2pm
+	if weekday == 6 && hour == 14 {
+		surveyMeetup(db, today)
+	}
+
+	// send chapter mtg surveys on sun at 4pm
+	if weekday == 0 && hour == 16 {
+		surveyChapterMeeting(db, today)
+	}
+
```
mdempsky: nit: Remove blank line. (Not a fan of blank lines before `}` lines.)

survey_mailer/survey_mailer.go
```diff
@@ -0,0 +1,204 @@
...
+		bodyHtml := bodyText
+		sendEmail(to, subject, bodyText, bodyHtml)
+	}
+}
+
+func sendEmail(to string, subject string, bodyText string, bodyHtml string) {
+	from := config.SurveyFromEmail
+	bodyHtml += `<br /><img src="https://adb.dxe.io/static/img/logo1.png" height="46" width="50">`
+	// EnvConfig uses the AWS credentials in the environment
+	// variables $AWS_ACCESS_KEY_ID nd $AWS_SECRET_KEY.
```
mdempsky: nit: "nd" should be "and".

survey_mailer/survey_mailer.go
```diff
@@ -0,0 +1,204 @@
...
+	"github.com/dxe/adb/model"
+	"github.com/jmoiron/sqlx"
+	"github.com/sourcegraph/go-ses"
+)
+
+func sendMissingEmail(eventName string, attendees []string) {
+	if len(attendees) > 0 {
+		to := config.SurveyMissingEmail
+		subject := "Missing emails for survey: " + eventName
+		bodyText := "The following people did not receive a survey for this protest due to not having a valid email address: "
```
mdempsky: nit: Should "protest" should probably be "event"?

survey_mailer/survey_mailer.go
```diff
@@ -0,0 +1,204 @@
...
+}
+
+func sendEmail(to string, subject string, bodyText string, bodyHtml string) {
+	from := config.SurveyFromEmail
+	bodyHtml += `<br /><img src="https://adb.dxe.io/static/img/logo1.png" height="46" width="50">`
+	// EnvConfig uses the AWS credentials in the environment
+	// variables $AWS_ACCESS_KEY_ID nd $AWS_SECRET_KEY.
+	_, err := ses.EnvConfig.SendEmailHTML(from, to, subject, bodyText, bodyHtml)
+	if err != nil {
+		log.Printf("Error sending email: %s\n", err)
```
mdempsky: I'd return the error so that the caller can handle it more intelligently. E.g., bulkSendEmails could not which send attempts fail, and include it in the `sendMissingEmail` alert; while `sendMissingEmail` could report it to the log.

survey_mailer/survey_mailer.go
```diff
@@ -0,0 +1,204 @@
...
+	_, err := model.UpdateEventSurveyStatus(db, model.Event{
+		ID:         eventId,
+		SurveySent: 1,
+	})
+	if err != nil {
+		log.Println("ERROR:", err)
+	}
+}
+
+func surveyMeetup(db *sqlx.DB, queryDate string) {
```
mdempsky: These three surveyFoo functions seem very similar. I'd try to write them as a single `surveyEvent` function that parameterizes the different conditions.

model/event.go
```diff
@@ -362,6 +373,41 @@ WHERE
...
 }
 
+func UpdateEventSurveyStatus(db *sqlx.DB, event Event) (eventID int, err error) {
+	tx, err := db.Beginx()
+	if err != nil {
+		return 0, errors.Wrap(err, "failed to update event")
+	}
+	// Error out if the event doesn't exist.
+	var eventCount int
+	err = tx.Get(&eventCount, `SELECT count(*) FROM events WHERE id = ?`, event.ID)
```
mdempsky: Doesn't `UPDATE` return the number of rows affected? Seems like having an extra query like this is unnecessary.

jakehobbs: Yeah, it does seem unnecessary for this. I had just borrowed it from updateEvent().

survey_mailer/survey_mailer.go
jakehobbs: Do you think it might be better to have this run every minute to check what time it is, then only have the other functions run when the time is exactly (to the minute) 9:00am, 2pm, or 4pm?

Alternatively, we could just have it run every 30 minutes, then there are two chances for it to run if something goes wrong. (And duplicates wouldn't be sent because things get marked as "sent" in the database.)

survey_mailer/survey_mailer.go
jakehobbs: Oo... Good idea. Will try that.

survey_mailer/survey_mailer.go
jakehobbs: Was thinking that would be a good idea, but wasn't sure the best way to implement it. Will do some more digging.

survey_mailer/survey_mailer.go
mdempsky: Thinking about it some more, it would probably be best to use a package like https://godoc.org/github.com/robfig/cron.

However, another thing to consider would be if there's a way to generalize the logic further, so you don't even have to do different logic at fixed schedules.

E.g., the real logic here seems to be that you want to process all events that haven't been processed yet, but you want to wait until a certain amount of time until after the event.

The former is already encoded in the database already. Could you check for the latter by looking for events where the time was at least N hours ago?

survey_mailer/survey_mailer.go
jakehobbs: I think I figured it out.

mdempsky: It looks like the main differences are the model.GetEventOptions, the description used for the bodyText/bodyHtml, and the URL. I'd think those should all be easily made into function parameters.

If having a bunch of function parameters becomes too unwieldy, you can put them into a struct type so that field keys (i.e., writing `Foo: x, Bar: y`) makes it clearer what each value means.

survey_mailer/survey_mailer.go
jakehobbs: Hmmm. Yes, it would work to just send surveys ~6 hours after events are created. Currently we aren't storing event time in the database though, so I'd have to add that first. And I'm slightly worried that doing so might create new issues.

An easier solution might be to just always send surveys the day after events. I don't think anyone would mind that.

mdempsky: Ah, forgot we only have date, not time.

survey_mailer/survey_mailer.go
jakehobbs: This is implemented in next commit.

survey_mailer/survey_mailer.go
jakehobbs: > An easier solution might be to just always send surveys the day after events. I don't think anyone would mind that.

I use this implementation in the next commit.

survey_mailer/survey_mailer.go
jakehobbs: Please review latest commit. Not sure if I did it the best way or not. It's my first time trying something like this in go.

model/event.go
```diff
@@ -219,7 +226,11 @@ WHERE
 	for _, a := range allAttendance {
 		i := eventIDToIndex[a.EventID]
 		events[i].Attendees = append(events[i].Attendees, a.ActivistName)
-		events[i].AttendeeEmails = append(events[i].AttendeeEmails, a.ActivistEmail)
+		if a.ActivistEmail != "" {
```
mdempsky: I don't think this is a good idea. Before we had the guarantee that Attendees[j] and AttendeeEmails[j] referred to the same activist, but now this will be lost if an attendee in the middle of the list is missing their email.

I think it's better for the survey mailer logic to handle keeping track of attendees who we don't have an email for and thus couldn't survey.

survey_mailer/survey_mailer.go
```diff
@@ -0,0 +1,198 @@
...
+		linkParam := ""
+		if surveyOptions.LinkParam == "name" {
+			linkParam = strings.Replace(event.EventName, " ", "+", -1)
+		}
+		if surveyOptions.LinkParam == "date" {
+			linkParam = event.EventDate.Format("2006-01-02")
+		}
+		// build body by replacing LINK_PARAM with the actual link param
+		bodyText := strings.Replace(surveyOptions.BodyText, "LINK_PARAM", linkParam, -1)
+		bodyHtml := strings.Replace(surveyOptions.BodyHtml, "LINK_PARAM", linkParam, -1)
```
mdempsky: String replacement into HTML is generally unsafe; e.g., the event name might contain <>.

This isn't a security issue since XSS doesn't apply to HTML email, and maybe we don't allow <> in event names anyway, but I'd probably at least put a `TODO` comment about applying proper escaping.

survey_mailer/survey_mailer.go
```diff
@@ -0,0 +1,198 @@
...
+}
+
+func bulkSendEmails(event model.Event, subject string, bodyText string, bodyHtml string) {
+	var sendingErrors []string
+	for _, recipient := range event.AttendeeEmails {
+		log.Println("Sending email to:", recipient)
+		// Send email
+		err := sendEmail(recipient, subject, bodyText, bodyHtml)
+		if err != nil {
+			sendingErrors = append(sendingErrors, err.Error())
```
mdempsky: Maybe change `err.Error()` to `recipient + " (" + err.Error() + ")"` (or vice versa) so that you'll know which address the error corresponds to.

model/event.go
jakehobbs: Fixed in next commit.

survey_mailer/survey_mailer.go
jakehobbs: Made a comment for now. Thanks.

jakehobbs: I am actually using the EscapeString function from here: https://golang.org/pkg/html/

Additionally, the ADB does not let you create event names with ```<```, ```>```, or ```&``` in the event name.

survey_mailer/survey_mailer.go
jakehobbs: Thanks. Did this in the next commit.

survey_mailer/survey_mailer.go
mdempsky: Marked as resolved but still says "nd"?

mdempsky:
Oh, you need to add another COPY line to the Dockerfile for the survey subdirectory. See models, mailinglist_sync, etc.

jakehobbs:
> LGTM w/ nit ("nd" -> "and")

Thanks. My bad missing that. I recall fixing it, but must have undid it and not noticed.

jakehobbs:
> Oh, you need to add another COPY line to the Dockerfile for the survey subdirectory. See models, mailinglist_sync, etc.
> 
> The Travis CI build should pass.

Thanks. Just did that. Will deploy in the morning.

jakehobbs:
PREREQS:

- [ ] Create mailing list for missing emails & sending errors to go to.
- [ ] Disable current GAS scripts on various spreadsheets: meetup, chapter mtg, protest.
- [ ] Add survey_sent col to prod events database (and set to 1 for previous day's events to prevent sending duplicates).
- [ ] Add new environment variables to docker instance (see readme).
- [ ] Remove surveys.sql from mysql-files and delete the csv files from www.

PR #60, opened 2019-12-12
remove unused model options

mdempsky:
The GetFoosJSON functions don't need to take Options parameters, since
they're not actually allowed to set any options anyway. Also,
eliminate the unused options (e.g., the "Name" field for circles and
working groups) and unimplemented options (e.g., most of the "User"
fields).

Low hanging fruit in trying to simplify the DB model to make better
sense of it.


jakehobbs:
Tested locally and looks good to me.

PR #61, opened 2019-12-12
simplify event_new and connection_new page rendering

mdempsky:
We don't actually incorporate the event details into the page using
the server-side templates anymore, so no point in querying the
database. Instead we can simply pass the event ID directly through.


jakehobbs:
Makes sense. Test locally.

PR #71, opened 2020-03-16
[WIP] Add CSRF protection to admin endpoints

liamkirsh:
The CSRF tokens that are generated can be reused for subsequent requests.

@mdempsky It would be useful to verify that these tokens can't be used across user accounts -- does that need to be tested in prod? E.g., I shouldn't be able to log in as myself, collect the CSRF base token and a request token, and then trick a user into submitting these values and have the server accept the request.

To generate the 32-byte random key for CSRF authentication, you can use `LC_ALL=C tr -dc '[:alnum:]' < /dev/urandom | head -c20`.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -20,8 +20,8 @@ var (
 	Route1 = mustGetenv("ROUTE_1", "/route1", true)
 	Route2 = mustGetenv("ROUTE_2", "/route2", true)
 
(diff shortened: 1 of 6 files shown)


mdempsky:
LGTM, but it needs the go.mod / go.sum files.

We'll also need to add CSRF_AUTH_KEY to the production environment before we can merge this.

PR #74, opened 2020-04-23
sync fb events

jakehobbs:


model/facebook.go
```diff
@@ -0,0 +1,138 @@
...
+	if len(events) == 0 {
+		// no pages in database
+		return nil, nil
+	}
+
+	return events, nil
+}
+
+func InsertFacebookEvent(db *sqlx.DB, event FacebookEventJSON, page FacebookPage) (nil, err error) { // we don't really need to return anything unless there's an error
+	tx, err := db.Beginx()
```
mdempsky, line 112: No need to wrap a single query in a transaction I think.

facebook_events/facebook_events.go
```diff
@@ -0,0 +1,83 @@
...
+	}
+	if resp.StatusCode != http.StatusOK {
+		panic(resp.StatusCode)
+	}
+	defer resp.Body.Close()
+	// read the response
+	body, err := ioutil.ReadAll(resp.Body)
+	// unmarshal the json data
+	data := model.FacebookResponseJSON{}
+	err = json.Unmarshal(body, &data)
```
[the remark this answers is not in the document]
jakehobbs: Ah, I'll change it to do that instead.

model/facebook.go
jakehobbs: Correct.

PR #89, opened 2020-11-06
added new API endpoints for discord bot

jakehobbs:
The idea behind this is to verify people's email address when they join the DxE Discord server, so that we can link their Discord ID to their activist record in the ADB.

The Discord bot (a separate node.js app) will message people when they join the server & hit these ADB endpoints to handle the user confirmation flow.

### DIFF
--- a/main.go
+++ b/main.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"context"
+	"crypto/rand"
 	"encoding/csv"
 	"encoding/json"
 	"fmt"
@@ -30,6 +31,7 @@ import (
 	"github.com/jmoiron/sqlx"
 	"github.com/justinas/alice"
 	"github.com/pkg/errors"
+	"github.com/sourcegraph/go-ses"
 	"github.com/urfave/negroni"
 )
 
@@ -73,6 +75,12 @@ func stripPort(ip string) string {
 		return strings.Split(ip, ":")[0]
 	}
 	return ip
+}
+
+func generateToken() string {
+	b := make([]byte, 32)
+	rand.Read(b)
+	return fmt.Sprintf("%x", b)
 }
 
 var sessionStore = sessions.NewCookieStore([]byte(config.CookieSecret))
@@ -238,6 +246,11 @@ func router() (*mux.Router, *sqlx.DB) {
 	// Authed Admin API for managing Users Roles
 	admin.Handle("/users-roles/add", alice.New(main.apiAdminAuthMiddleware).ThenFunc(main.UsersRolesAddHandler))
 	admin.Handle("/users-roles/remove", alice.New(main.apiAdminAuthMiddleware).ThenFunc(main.UsersRolesRemoveHandler))
+
+	// Discord API
+	router.Handle("/discord/status", alice.New(main.discordBotAuthMiddleware).ThenFunc(main.DiscordStatusHandler))
+	router.Handle("/discord/generate", alice.New(main.discordBotAuthMiddleware).ThenFunc(main.DiscordGenerateHandler))
+	router.HandleFunc("/discord/confirm/{id:[0-9]+}/{token:[a-zA-Z0-9]+}", main.DiscordConfirmHandler)
 
 	// Pprof debug routes
 	router.HandleFunc("/debug/pprof/", pprof.Index)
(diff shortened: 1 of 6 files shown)


main.go
```diff
@@ -1547,6 +1560,159 @@ func (c MainController) ListAllChapters(w http.ResponseWriter, r *http.Request)
 	writeJSON(w, chapters)
 }
 
+// testing this
```
jakehobbs, line 1563: I will remove this comment, as it's tested and works.

main.go
```diff
@@ -1547,6 +1560,159 @@ func (c MainController) ListAllChapters(w http.ResponseWriter, r *http.Request)
...
+
+		err := r.ParseForm()
+		if err != nil {
+			http.Error(w, http.StatusText(400), 400)
+			return
+		}
+
+		// check that discord auth matches (shared secret w/ discord bot)
+		discordAuth := r.PostFormValue("auth")
+		if discordAuth != config.DiscordSecret {
```
mdempsky, line 1575: At least theoretically, this is vulnerable to a timing attack. Might as well use  `subtle.ConstantTimeCompare([]byte(discordAuth), []byte(config.DiscordSecret)) != 1` instead though.

main.go
```diff
@@ -1547,6 +1560,159 @@ func (c MainController) ListAllChapters(w http.ResponseWriter, r *http.Request)
...
+		})
+		// TODO: assign roles in discord immediately once bot is ready to take requests
+		return
+	}
+
+	// render error page
+	renderPage(w, r, "discord", PageData{
+		PageName: "Error",
+		Data: map[string]interface{}{
+			"message": "There was a problem verifying your email. Please try again or contact " + template.HTML(`<a href="mailto:tech@dxe.io">tech@dxe.io</a>`) + ".",
```
jakehobbs, line 1710: I might move the "support email address" to the config.

main.go
```diff
@@ -1547,6 +1560,159 @@ func (c MainController) ListAllChapters(w http.ResponseWriter, r *http.Request)
...
+func (c MainController) DiscordStatusHandler(w http.ResponseWriter, r *http.Request) {
+	err := r.ParseForm()
+	if err != nil {
+		writeJSON(w, map[string]interface{}{
+			"status": "error parsing form",
+		})
+		return
+	}
+
+	discordUserID, err := strconv.Atoi(r.PostFormValue("id"))
```
mdempsky, line 1593: Check `err != nil`?

main.go
```diff
@@ -1547,6 +1560,159 @@ func (c MainController) ListAllChapters(w http.ResponseWriter, r *http.Request)
...
+	if err != nil {
+		panic(err.Error())
+	}
+
+	// trigger email to be sent w/ verification link
+	if config.DiscordFromEmail != "" && config.AWSAccessKey != "" && config.AWSSecretKey != "" && config.AWSSESEndpoint != "" {
+		subjectText := "Please verify your email"
+		// TODO: make nicer looking email
+		confirmLink := config.UrlPath + "/discord/confirm/" + strconv.Itoa(user.ID) + "/" + user.Token
+		bodyText := "Hello, Please click this link to verify your email address: " + confirmLink + " Cheers, The DxE Discord Bot"
```
mdempsky, line 1656: It would probably be good to add some newlines before and after the link so it's easier to click.

main.go
```diff
@@ -1547,6 +1560,159 @@ func (c MainController) ListAllChapters(w http.ResponseWriter, r *http.Request)
...
+	vars := mux.Vars(r)
+	id, err := strconv.Atoi(vars["id"])
+	if err != nil {
+		panic(err)
+	}
+	user.ID = id
+	user.Token = vars["token"]
+
+	// try to confirm user
+	_, err = model.ConfirmDiscordUser(c.db, user)
```
mdempsky, line 1684: Check `err != nil`?

model/discord.go
```diff
@@ -0,0 +1,60 @@
...
+	Email     string `db:"email"`
+	Token     string `db:"token"`
+	Confirmed bool   `db:"confirmed"`
+}
+
+type DiscordUserStatus string
+
+const (
+	NotFound  DiscordUserStatus = "not found"
+	Pending                     = "pending"
```
mdempsky, line 19: You need to put `DiscordUserStatus` for `Pending` and `Confirmed`. Because an explicit value was provided, the type doesn't get copied from the previous line.

model/discord.go
```diff
@@ -0,0 +1,60 @@
...
+		if users[0].Confirmed {
+			return Confirmed, nil
+		} else {
+			return Pending, nil
+		}
+	}
+	return NotFound, nil
+}
+
+func ConfirmDiscordUser(db *sqlx.DB, user DiscordUser) (bool, error) {
```
mdempsky, line 54: I'd probably return just `error`. The `bool` is redundant.

main.go
```diff
@@ -75,6 +77,12 @@ func stripPort(ip string) string {
 	return ip
 }
 
+func generateToken() string {
+	b := make([]byte, 32)
+	rand.Read(b)
```
mdempsky, line 82: I don't think this should ever fail, but it's worth checking. Panic is fine.

mdempsky (approved):

main.go
jakehobbs: This is just for testing everything right now. I'll make the email look nicer later.

PR #102, opened 2021-01-29
add events to eventbrite automatically

jakehobbs:
This is mostly done, but I still need to add functionality for updating existing Eventbrite events when changes are detected on Facebook.


event_sync/eventbrite_create.go
```diff
@@ -0,0 +1,335 @@
...
+		Lng        string `json:"longitude"`
+	}
+	type venue struct {
+		Name    string  `json:"name"`
+		Address address `json:"address"`
+	}
+	type request struct {
+		Venue venue `json:"venue"`
+	}
+	type response struct {
```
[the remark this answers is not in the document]
mdempsky: Alternatively, you're doing the whole encode+http.Post+check-error+decode stuff is pretty common. You might try refactoring out a helper function like:

```
func callAPI(path string, req, resp *interface{}) error {
    // JSON encode req to a bytes.Body
    // do an http.Post using path, "application/json", and body
    // check for HTTP errors
    // JSON decode resp.Body into resp
}
```

event_sync/images.go
```diff
@@ -0,0 +1,182 @@
...
+	}
+
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		return image, errors.New("failed to get image from Facebook. Status: " + strconv.Itoa(resp.StatusCode))
+	}
+
+	image.Buffer = make([]byte, resp.ContentLength)
+	_, err = io.ReadFull(resp.Body, image.Buffer)
```
mdempsky, line 48: FYI, you could also do `image.Buffer, err = ioutil.ReadAll(resp.Body)` if you wanted the code to be a little simpler, but this code is actually fine and more efficient, because it knows the size of the buffer in advance so it only allocates once.

event_sync/images.go
```diff
@@ -0,0 +1,182 @@
...
+	err = json.NewDecoder(resp.Body).Decode(&data)
+	if err != nil {
+		return data, err
+	}
+	return data, nil
+}
+
+func uploadImageToEventbrite(image Image, data EventbriteUploadData) error {
+	// create a buffer to store the form data that's needed for S3
+	buf := new(bytes.Buffer)
```
mdempsky, line 83: `var buf bytes.Buffer` is a bit more idiomatic, I think. But this is fine too.

event_sync/images.go
```diff
@@ -0,0 +1,182 @@
...
+	}
+
+	// upload it
+	resp, err := http.Post(data.UploadURL, w.FormDataContentType(), buf)
+	if err != nil {
+		return err
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != 204 {  // This endpoint returns 204 instead of 200.
```
mdempsky, line 121: Could use http.StatusNoContent, but 204 is fine too.

mdempsky (approved):

event_sync/eventbrite_create.go
```diff
@@ -0,0 +1,335 @@
+package event_sync
+
+import (
+	"bytes"
+	"encoding/json"
+	"errors"
+	"github.com/dxe/adb/model"
```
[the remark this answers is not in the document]
jakehobbs: Thanks. I just enabled goimports.

event_sync/event_sync.go
```diff
@@ -112,82 +114,179 @@ func syncFacebookEvents(db *sqlx.DB) {
...
+}
+
+// This function will always need to match all existing events since the Facebook Sync will wipe out the data.
+func addEventbriteDataToExistingEvents(db *sqlx.DB, ebEvents []model.EventbriteEventJSON) error {
+	if len(ebEvents) > 0 {
+		// loop through events
+		for _, ebEvent := range ebEvents {
+				err := model.AddEventbriteDetailsToEventByName(db, ebEvent)
+				if err != nil {
+					return err
```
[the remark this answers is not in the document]
jakehobbs: There's not any concern for that IMO.

event_sync/images.go
jakehobbs: I'll use that code instead. I spent a fair amount of time trying to figure out how to get this to work, so your simpler option seems better since the efficiency here doesn't matter much.

event_sync/eventbrite_create.go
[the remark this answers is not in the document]
jakehobbs: Good thinking. That should simplify things a bit.

event_sync/utils.go
```diff
@@ -0,0 +1,48 @@
...
+func callAPIGet(path string, resp interface{}) error {
+	response, err := http.Get(path)
+	if err != nil {
+		return err
+	}
+	defer response.Body.Close()
+	if response.StatusCode != http.StatusOK {
+		return errors.New("GET request failed. Status: " + strconv.Itoa(response.StatusCode))
+	}
+	err = json.NewDecoder(response.Body).Decode(&resp)
```
mdempsky, line 45: return the error?

event_sync/utils.go
```diff
@@ -0,0 +1,48 @@
...
+	if err != nil {
+		return err
+	}
+	defer response.Body.Close()
+
+	if response.StatusCode != http.StatusOK {
+		return errors.New("POST request failed. Status: " + strconv.Itoa(response.StatusCode))
+	}
+
+	err = json.NewDecoder(response.Body).Decode(&resp)
```
mdempsky, line 28: I would just directly return the call expression, rather than assigning to `err` and the `if` statement.

event_sync/utils.go
```diff
@@ -0,0 +1,48 @@
...
+
+import (
+	"bytes"
+	"encoding/json"
+	"errors"
+	"net/http"
+	"strconv"
+)
+
+func callAPIPost(path string, req, resp interface{}) error {
```
mdempsky, line 11: nit: I might shorten to just `postAPI` (and `getAPI` below), but I don't feel strongly.

main.go
```diff
@@ -1433,34 +1432,32 @@ func (c MainController) ListFBEventsHandler(w http.ResponseWriter, r *http.Reque
...
+	var startTime, endTime time.Time
+	var err error
+
+	if startTimeStr, ok := params["start_time"]; ok {
+		startTime, err = time.Parse(time.RFC3339, startTimeStr[0])
+		if err != nil {
+			writeJSON(w, map[string]string{
 				"error": "start_time format incorrect",
 			})
-			return
```
mdempsky, line 1446: Why remove the returns after writeJSON?

event_sync/utils.go
jakehobbs: whoops

main.go
jakehobbs: That was a mistake. Thanks for catching.

PR #109, opened 2021-03-21
Add adb-forms golang form processor

AlexisDeschamps:
Add `processor_go` from https://github.com/dxe/adb-forms.

- Copy the files from the other repo
- Run the form processor from `main.go`
- Add `.env` to `.gitignore`

It seems like everything works without any issues based on local testing.


jakehobbs (commented):
I would suggest sharing some of the config (at least things like the mysql connection & smtp settings) with the ADB's main config. Looks good merge after that!

jakehobbs (approved):

PR #150, opened 2021-05-16
make the output of `make fmt` pass on master

AlexisDeschamps:
- pin version of `prettier` to `2.3.0` to make the CI and local development use the same version
- add the output of `make fmt`


jakehobbs:
LGTM

PR #151, opened 2021-05-23
Pin all node packages

AlexisDeschamps:
Why: avoid unexpected differences between different environments
What: pin all node packages to their current minimum version


jakehobbs:
Looks good to me.

jakehobbs (approved):

PR #152, opened 2021-05-24
Do not exit on OIDC provider construction failures

AlexisDeschamps:
Why: be able to work on more parts of the application while offline
What: do not exit on OIDC provider construction failures. Log a warning instead.

### DIFF
--- a/main.go
+++ b/main.go
@@ -440,7 +440,7 @@ func getUserFromContext(ctx context.Context) model.ADBUser {
 var verifier = func() *oidc.IDTokenVerifier {
 	provider, err := oidc.NewProvider(context.Background(), "https://accounts.google.com")
 	if err != nil {
-		log.Println("failed to create OIDC provider (no Internet connection)?: ", err)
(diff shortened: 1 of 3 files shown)


main.go
```diff
@@ -440,7 +440,8 @@ func getUserFromContext(ctx context.Context) model.ADBUser {
 var verifier = func() *oidc.IDTokenVerifier {
 	provider, err := oidc.NewProvider(context.Background(), "https://accounts.google.com")
 	if err != nil {
-		log.Fatal(err)
+		log.Println("WARNING: failed to construct OIDC provider (no Internet connection?). Some features may not work: ", err)
```
jakehobbs, line 443: Small nit: Println automatically adds spaces between each operand, so you don't need a space after the colon (unless you were trying to print two spaces).

jakehobbs (approved):
Works great. Thanks.

main.go
AlexisDeschamps: Thanks for the tip, fixed it

PR #197, opened 2023-05-02
Geocoding on activist update

jefftienchow:
Attempt at geocoding addresses with google API on activist update

Tested locally on a variety of addresses. Google Geocoding API has a bit undefined behavior when entering bogus addresses but seems to work well with real addresses. 

Not sure if I should include package-lock.json changes in my commits?

### DIFF
--- a/model/activist.go
+++ b/model/activist.go
@@ -5,14 +5,12 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
-	"net/http"
-	"net/url"
+
 	"regexp"
 	"strconv"
 	"strings"
 	"time"
 
-	"github.com/dxe/adb/config"
 	"github.com/dxe/adb/mailing_list_signup"
 
 	"github.com/go-sql-driver/mysql"
@@ -444,6 +442,7 @@ type GetActivistOptions struct {
 	UpcomingFollowupsOnly bool   `json:"upcoming_followups_only"`
 }
 
+// struct for geocoding API: https://developers.google.com/maps/documentation/geocoding/overview
 type GeocodeResponse struct {
 	Results []struct {
 		Geometry struct {
@@ -1145,17 +1144,18 @@ func UpdateActivistData(db *sqlx.DB, activist ActivistExtra, userEmail string) (
 	// if in the Bay Area, add to mailing list if needed
 	// TODO: make this work for other chapters too, maybe based on Chapters table mailing list info?
 	// or maybe just passing chapter Name or ID to the signup service?
-	if origActivist.ChapterID == 47 {
-		activistAddressChanged := activist.StreetAddress != origActivist.StreetAddress ||
-			activist.City != origActivist.City ||
-			activist.State != origActivist.State
-		activistInfoChanged := activistAddressChanged ||
-			activist.Location != origActivist.Location ||
-			activist.Name != origActivist.Name ||
+	if origActivist.ChapterID == 1 {
+		geoInfoChanged := activist.City != origActivist.City ||
+			activist.State != origActivist.State || 
+			activist.StreetAddress != origActivist.StreetAddress
+		mailingListInfoChanged := activist.Name != origActivist.Name ||
 			activist.Email != origActivist.Email ||
 			activist.Phone != origActivist.Phone ||
+			activist.Location != origActivist.Location ||
+			activist.City != origActivist.City ||
+			activist.State != origActivist.State ||
 			activist.ActivistLevel != origActivist.ActivistLevel
-		if activistInfoChanged && activist.Email != "" {
+		if mailingListInfoChanged && activist.Email != "" {
 			signup := mailing_list_signup.Signup{
 				Source:        "adb",
 				Name:          activist.Name,
@@ -1172,22 +1172,11 @@ func UpdateActivistData(db *sqlx.DB, activist ActivistExtra, userEmail string) (
 				fmt.Println("ERROR updating activist on mailing list:", err.Error())
 			}
 		}
-		if activistAddressChanged && activist.StreetAddress != "" && activist.City != "" && activist.State != "" {
-			full_address := url.QueryEscape(activist.StreetAddress + " " + activist.City + " " + activist.State)
-			request := "https://maps.googleapis.com/maps/api/geocode/json?address=" + full_address + "&key=" + config.GooglePlacesAPIKey
-			resp, err := http.Get(request)
-			if err != nil {
-				fmt.Println("Error geocoding activist location", err)
-			}
-			defer resp.Body.Close()
-			var geocode_response GeocodeResponse
-			json.NewDecoder(resp.Body).Decode(&geocode_response)
-			if len(geocode_response.Results) == 0 {
-				fmt.Printf("No geocoding results found for address %v. Not updating Lat and Lng", full_address)
-			} else {
-				activist.Lng = geocode_response.Results[0].Geometry.Location.Lng
-				activist.Lat = geocode_response.Results[0].Geometry.Location.Lat
-			}
+		if geoInfoChanged && activist.StreetAddress != "" && activist.City != "" && activist.State != "" {
+			location := geoCodeAddress(activist.StreetAddress, activist.City, activist.State)
+			activist.Lng = location.Lng
+			activist.Lat = location.Lat
+			fmt.Println("updated:", activist.Lng, activist.Lat)
 		}
 	}
 
--- a/model/activist.go
+++ b/model/activist.go
@@ -1144,7 +1144,7 @@ func UpdateActivistData(db *sqlx.DB, activist ActivistExtra, userEmail string) (
 	// if in the Bay Area, add to mailing list if needed
 	// TODO: make this work for other chapters too, maybe based on Chapters table mailing list info?
 	// or maybe just passing chapter Name or ID to the signup service?
-	if origActivist.ChapterID == 1 {
+	if origActivist.ChapterID == 47 {
 		geoInfoChanged := activist.City != origActivist.City ||
 			activist.State != origActivist.State || 
 			activist.StreetAddress != origActivist.StreetAddress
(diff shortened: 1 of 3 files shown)


jakehobbs:
Thanks for this! I will review later today.

alexsapps:
> Not sure if I should include package-lock.json changes in my commits?

Yes, maintaining package-lock.json changes in version control ensures reproducible builds. package.json only specifies a range of versions allowed to build the project, not exact versions.

model/activist.go
```diff
@@ -1131,12 +1146,14 @@ func UpdateActivistData(db *sqlx.DB, activist ActivistExtra, userEmail string) (
 	// TODO: make this work for other chapters too, maybe based on Chapters table mailing list info?
 	// or maybe just passing chapter Name or ID to the signup service?
 	if origActivist.ChapterID == 47 {
-		activistInfoChanged := activist.Name != origActivist.Name ||
+		activistAddressChanged := activist.StreetAddress != origActivist.StreetAddress ||
```
alexsapps, line 1149: looks like StreetAddress is not used by the mailing list, so this should not trigger a call to update the mailing list.

model/activist.go
```diff
@@ -1131,12 +1146,14 @@ func UpdateActivistData(db *sqlx.DB, activist ActivistExtra, userEmail string) (
...
+			activist.City != origActivist.City ||
+			activist.State != origActivist.State
+		activistInfoChanged := activistAddressChanged ||
+			activist.Location != origActivist.Location ||
```
alexsapps, line 1153: should location be considered in activistAddressChanged instead? it look like it contains the Zip which is used by the mailing list.

but then Zip is not used by the geo API, so maybe this logic should not be shared between geo api logic and mailing list logic, in which case using names like "isMailingListInfoChanged" and "isGeoInfoChanged" would better convey the intent.

model/activist.go
```diff
@@ -1155,7 +1172,23 @@ func UpdateActivistData(db *sqlx.DB, activist ActivistExtra, userEmail string) (
 				fmt.Println("ERROR updating activist on mailing list:", err.Error())
 			}
 		}
-		fmt.Println("Updated: ", activist.Location, activist.City, activist.State)
+		if activistAddressChanged && activist.StreetAddress != "" && activist.City != "" && activist.State != "" {
```
alexsapps, line 1175: please split the body of this 'if' into a separate function (e.g. called `GetCoordinatesFromAddress`) in a separate file (single responsibility principle). i suggest the function take a single string parameter for the address and return the location and any error.

additionally, but optionally, the 'if' itself could be its own function called `updateActivistCoordinates` which would allow early returns in case of error rather than nesting the happy path inside of an `if` block.

model/activist.go
```diff
@@ -1155,7 +1172,23 @@ func UpdateActivistData(db *sqlx.DB, activist ActivistExtra, userEmail string) (
...
+			full_address := url.QueryEscape(activist.StreetAddress + " " + activist.City + " " + activist.State)
+			request := "https://maps.googleapis.com/maps/api/geocode/json?address=" + full_address + "&key=" + config.GooglePlacesAPIKey
+			resp, err := http.Get(request)
+			if err != nil {
+				fmt.Println("Error geocoding activist location", err)
```
alexsapps, line 1180: if there is an error might be better to break out of this `if` somehow (or return early if you split this into another function)

model/activist.go
```diff
@@ -441,6 +444,18 @@ type GetActivistOptions struct {
 	UpcomingFollowupsOnly bool   `json:"upcoming_followups_only"`
 }
 
+type GeocodeResponse struct {
```
alexsapps, line 447: please a comment indicating that this structure corresponds to Google's geo coding API with a link to that API's documentation

alexsapps (changes requested):

model/activist.go
jakehobbs: On this note, I think we'd probably prefer to silently fail (& log it) but still go through w/ saving the activist in the db if the gecocoding service fails for w/e reason.

I feel it's pretty important that activist data is still able to be updated even if the geocoding service is failing.

jakehobbs:
I took a brief look over this and p much agree w/ what Alex pointed out. Left one comment on one of his comments, but I don't have anything else to add other than that.

jakehobbs (commented):
I took a brief look over this and p much agree w/ what Alex pointed out. Left one comment on one of his comments, but I don't have anything else to add other than that.

model/activist.go
jefftienchow: Yeah I don't think zip is used for geocoding so I'll do the latter

model/activist.go
jakehobbs: Thinking about this further... maybe it would be ideal to just fire off a goroutine here that handles the things that need to happen in the background, such as updating mailing list and coords? B/c we don't really want the user to have to wait around for those things to happen before the activist is saved in the db. (This was a bit of a flaw in the original logic for updating the mailing list.)

alexsapps: probably the coords should still be looked synchronously because the result is used in the database update? if so, could the signup be made async in a separate PR so Jeff's change isn't held up?

just wondering, is the async idea to make it faster when an admin of ADB clicks save to update a user? did you notice the adb signup was slow?

maybe whatever is simplest/easiest is best for this probably rarely used feature. but some possible improvements mainly as thought experiments:
* it would be nice to know if the signup operations fails, maybe by an email notification to our team, perhaps directly or by monitoring error logs.
* it would be nice to ensure async operations complete even if the web server container is requested to shut down or even if an error in the main function causes it to return and shut down. perhaps there would be a work queue that the signup gets added to, and the main function can wait for the queue on the last line before exiting.

jakehobbs: Fair points. Yeah, I was thinking the main reason would be so it saves faster, but I that might be a bit of an over optimization for now. I def won't block on it. 

We have log drains in AWS and can set up Cloudwatch alerts for whatever.

The signup service has its own queue. That being said, I do think the background jobs that run in ADB could use a huge overhaul at some point if we have bandwidth. Right now they only run within a single instance of the app (via an environment variable that enables them). It would be great to implement a more robust job system.

jakehobbs: Maybe walking through how the signup service works at the next meeting would be useful. A lot of stuff feeds in and out of it, and would be good for others to know how it works.

model/activist.go
```diff
@@ -1131,14 +1145,17 @@ func UpdateActivistData(db *sqlx.DB, activist ActivistExtra, userEmail string) (
...
 	// or maybe just passing chapter Name or ID to the signup service?
 	if origActivist.ChapterID == 47 {
-		activistInfoChanged := activist.Name != origActivist.Name ||
+		geoInfoChanged := activist.City != origActivist.City ||
+			activist.State != origActivist.State ||
+			activist.StreetAddress != origActivist.StreetAddress
+		mailingListInfoChanged := activist.Name != origActivist.Name ||
 			activist.Email != origActivist.Email ||
 			activist.Phone != origActivist.Phone ||
 			activist.Location != origActivist.Location ||
```
alexsapps, line 1154: would be nice to move this down a couple lines underneath the state != state check to keep things in the same order that they are used later on, to more easily verify correctness at a glance.

model/activist.go
```diff
@@ -1131,14 +1145,17 @@ func UpdateActivistData(db *sqlx.DB, activist ActivistExtra, userEmail string) (
 	// TODO: make this work for other chapters too, maybe based on Chapters table mailing list info?
 	// or maybe just passing chapter Name or ID to the signup service?
 	if origActivist.ChapterID == 47 {
-		activistInfoChanged := activist.Name != origActivist.Name ||
+		geoInfoChanged := activist.City != origActivist.City ||
```
alexsapps, line 1148: please declare this where it is used, just above `if geoInfoChanged && ...`

model/geo_helpers.go
```diff
@@ -0,0 +1,34 @@
...
+		return nil
+	}
+	defer resp.Body.Close()
+	var geocode_response GeocodeResponse
+	json.NewDecoder(resp.Body).Decode(&geocode_response)
+	if len(geocode_response.Results) == 0 {
+		fmt.Printf("No geocoding results found for address %v. Not updating Lat and Lng", full_address)
+		return nil
+	} else {
+		return &Location{Lat: geocode_response.Results[0].Geometry.Location.Lat, Lng: geocode_response.Results[0].Geometry.Location.Lng}
```
alexsapps, line 32: If you wouldn't mind checking if Results[0] and Geometry and Location are non-nil, I think that could prevent a segfault. Even if the API always returns non-nil now, I worry this could change and we wouldn't want the server to crash - just log an error and carry on.

alexsapps (changes requested):
Thanks for the updates! Just got a few more suggestions.

model/geo_helpers.go
jefftienchow: I'm a beginner at Go, but from my understanding this wouldn't be possible? Constructing a GeocodeResponse from json that doesn't contain these fields would just result in the default values being included for lat and lng, which would be 0 and 0

jefftienchow: https://stackoverflow.com/questions/50597713/check-if-struct-field-is-empty

model/geo_helpers.go
```diff
@@ -0,0 +1,47 @@
...
+	Status string `json:"status"`
+}
+
+func geoCodeAddress(streetAddress string, city string, state string) *Location {
+	full_address := url.QueryEscape(streetAddress + " " + city + " " + state)
+	request := "https://maps.googleapis.com/maps/api/geocode/json?address=" + full_address + "&key=" + config.GooglePlacesAPIKey
+	resp, err := http.Get(request)
+	if err != nil {
+		fmt.Println("Error geocoding activist location", err)
+		return nil
```
alexsapps, line 36: optional: return the error (and print it in the caller) to allow callers to handle errors as needed, thereby making this function more reusable

model/geo_helpers.go
```diff
@@ -0,0 +1,47 @@
...
+	}
+	defer resp.Body.Close()
```
alexsapps, line 38: consider adding a blank line between the HTTP fetching logic and GeocodeResponse parsing logic.

model/geo_helpers.go
alexsapps: Oh got it. My bad :) I'm a Go beginner too. So a struct inside struct has a default value of the 0 value of the inner struct, as opposed to a struct *pointer* inside a struct which would be nil.

alexsapps (approved):
Looks good to me. Let's see if Jake has any suggestions since it's the first PR in this codebase for either of us, I think.

Can you select "squash" merge strategy rather than "merge"? Otherwise it will clutter the main branch log since the many commits in this branch are all for one logical change. If it lets you type a commit message, perhaps use something slightly more precise like "Geocode activist address on update"

model/geo_helpers.go
jakehobbs: yeah i agree returning the error would be good as best practice, then the caller could just choose to ignore it if you don't want to actually handle it.

jakehobbs (approved):

PR #211, opened 2024-05-30
Update logic for coloring chapters' last action

alexsapps:


frontend/ChapterList.vue
```diff
@@ -871,10 +871,10 @@ export default Vue.extend({
       if (time.isValid()) {
         c = 'is-danger';
       }
-      if (time.isAfter(dayjs().add(-2, 'month'))) {
+      if (time.isAfter(dayjs().add(-58, 'day'))) {
```
jakehobbs, line 874: yuck forgot we were still using dayjs in this 🤣

fine thought for now obvs

jakehobbs (approved):

frontend/ChapterList.vue
alexsapps: lol. noted dayjs is yuck.

jakehobbs: not as bad as moment. i didn't realize it's still being updated, but it is. i prefer date-fns these days though, and it has a smaller bundle size.

PR #214, opened 2024-12-28
scaffold react app

jakehobbs:


alexsapps (approved):
If you are up to it, would be nice for the commit message to include the command you used to create this scaffolding so we can see any options that vary from the default and cross reference our code with the output of that command for future debugging needs. But not super necessary.

Makefile
```diff
@@ -1,8 +1,10 @@
 .PHONY: run_all run watch test clean prod_build deps dev_db fmt
 
-# Runs the application.
+# Runs the application (builds Vue.js files, starts Next.js dev server, starts Go server).
 run_all:
-	cd frontend && npm run dev-build
+	. ~/.nvm/nvm.sh && \
+    (cd frontend && nvm install 16 && npm run dev-build) && \
```
jakehobbs, line 6: This is kinda annoying. But we have some super old deps in the Vue app that break if we go past Node 16. But we need at least 18 to use the latest React.

frontend/ActivistList.vue
```diff
@@ -1903,8 +1903,9 @@ export default Vue.extend({
       height: 500,
       columns: getColumnsForChapter(this.chapter, this.view),
       lastEventDateFrom:
-        this.view === ('all_activists' || 'leaderboard') ? initialDateFromValue() : '',
-      lastEventDateTo: this.view === ('all_activists' || 'leaderboard') ? initialDateToValue() : '',
+        this.view === 'all_activists' || this.view === 'leaderboard' ? initialDateFromValue() : '',
```
jakehobbs, line 1906: bug fix

frontend/ChapterList.vue
```diff
@@ -633,7 +633,7 @@ export default Vue.extend({
         this.hideModal();
       }
 
-      this.currentChapter = { ...chapter };
+      this.currentChapter = !chapter ? ({} as Chapter) : { ...chapter };
```
jakehobbs, line 636: these were just some small changes to make TS compiler happy

frontend/UserList.vue
```diff
@@ -136,6 +136,7 @@ interface User {
   email: string;
   roles: string[];
   chapter_id: number;
+  disabled: boolean;
```
jakehobbs, line 139: this should have always been here, but it was missing

frontend/nav.json
```diff
@@ -0,0 +1,141 @@
+{
```
jakehobbs, line 1: i moved the menu data here in anticipation of creating a navbar component in react, so that we can share the data b/w react & vue.

server/src/main.go
```diff
@@ -229,6 +247,7 @@ func router() (*mux.Router, *sqlx.DB) {
 
 	// Unauthed API (public)
 	router.Handle("/health", alice.New(main.corsAllowGetMiddleware).ThenFunc(main.HealthStatusHandler))
+	router.Handle("/static_resources_hash", alice.New(main.corsAllowGetMiddleware).ThenFunc(main.StaticResourcesHashHandler))
```
jakehobbs, line 250: not the biggest fan of this, but i think it's fine for now

frontend-v2/src/components/VueNavbar.tsx
```diff
@@ -0,0 +1,46 @@
+import { useSession } from "@/hooks/use-session";
+import Script from "next/script";
+import { useStaticResourceHash } from "@/hooks/use-static-resource-hash";
+
+// Allows the Vue AdbNav component to be used within the React app
+// for more seamless UX. Once most pages are rebuilt in
```
alexsapps, line 6: could you explain the seamless ux part? what is the implied alternative that would be less seamless?

also wondering how the vue navbar works in this context. do its links go to the vue app or stay in the react app? does it load vue components in react?

Makefile
alexsapps: that would be useful to add as a comment

// As of January 2025, upgrading past Node 16 breaks old Vue dependencies, and Node 18 is required to use the latest version of React.

frontend-v2/README.md
alexsapps, line 1: Could you explain at a high level how this migration is set up? The context would help me review.

Some questions I have:
* Does this share auth cookies with the Vue frontend or do they sign into both apps?
* Is the app intended to be served from the same subdomain or another one?
* How are Vue components loaded in? How does the navigation work? Can users see both versions of pages at different URLs or do new pages replace old pages?

frontend-v2/src/hooks/use-static-resource-hash.ts
```diff
@@ -0,0 +1,20 @@
+import { useQuery } from "@tanstack/react-query";
+import ky from "ky";
+import { z } from "zod";
+
+const StaticResourcesHashResp = z.object({
+  hash: z.string(),
+});
+
+// This is only used for Vue components and should eventually be removed.
```
alexsapps, line 9: in case we need to look back a long time from now (even though it's more temporary than other code), could you explain what this does too? if it's some technique you found online, could be helpful to link to the source.

frontend/ChapterList.vue
alexsapps: is making it an optional parameter also necessary for the ts compiler? if the old code was working, it seems to imply that the args passed were always non-nullish.
and is this not only necessary when the parameter is optional?

but since this code should hopefully be getting deleted soon, feel free to ignore if it takes non-trivial looking into :)

alexsapps (changes requested):

frontend/ChapterList.vue
jakehobbs: We are calling it in places with only 1 param, so it should be optional unless we change that.

frontend-v2/src/components/ContentWrapper.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+const contentWrapperClass = {
+  sm: "lg:max-w-screen-sm",
+  md: "lg:max-w-screen-md",
+  lg: "lg:max-w-screen-lg",
+  xl: "lg:max-w-screen-xl",
+  "2xl": "lg:max-w-screen-2xl",
+};
+
+export const ContentWrapper = (props: {
```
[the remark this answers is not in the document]
jakehobbs: sure, it just does the same thing as the "body wrapper" from the vue app

frontend-v2/src/components/VueNavbar.tsx
jakehobbs: seamless b/c the navbar looks the same on all pages (vue and react), and doesn't require us changing the code in 2 places. as opposed to alternative of having 2 navbars that look different when you navigate b/w the react pages & the vue pages.

the vue navbar within the react app goes to the vue pages still like it always has.

frontend-v2/README.md
jakehobbs: auth is handled the same way as always. (it's actually mostly handled server-side and in vanilla js, not within the vue components.)

the react app is served at all `/v2/*` paths for now (via a proxy in local dev so the next.js dev server can properly run, then via static files in the prod build).

vue pages & navigation work exactly the same as they always have. just if a user goes to a `/v2` page, it loads the react app. one a new page is ready, it will replace an old page (incrementally).

frontend-v2/src/hooks/use-static-resource-hash.ts
jakehobbs: using it like we use it in other places in the codebase, just added a new endpoint to fetch it w/o doing the go template SSR stuff since the react code is not doing any SSR.

frontend-v2/README.md
alexsapps: oh okay. sounds like /v2/* starts out as basically the same as the original app, although it is an independent frontend except for auth and any other vanilla JS, and its pages will be incrementally converted to react, but the vue pages will remain available at their current locations.

frontend-v2/src/components/ui/button.tsx
alexsapps, line 1: did we hand-write all this, or if there are any instructions to generate more of these maybe we could explain in a /ui/README.md with just a link to whatever like https://www.radix-ui.com/ ?

frontend-v2/src/lib/utils.ts
```diff
@@ -0,0 +1,6 @@
+import { clsx, type ClassValue } from "clsx";
+import { twMerge } from "tailwind-merge";
+
+export function cn(...inputs: ClassValue[]) {
```
alexsapps, line 4: what does CN stand for? would be nice to have a comment that introduces this function without having to infer from usage / dig into the dependencies' docs, however briefly. no need to explain everything completely since it's obviously just wrapping something else.

alexsapps (approved):
Approving in case we decide to add comments in a separate PR or if you want me to add the comments.

frontend-v2/src/components/ui/button.tsx
jakehobbs: the components in the `ui` directory are initially generated by shadcn. will note in the readme.

jakehobbs: https://ui.shadcn.com/docs

frontend-v2/src/lib/utils.ts
jakehobbs: this util is generated by shadn. i'll add a link to an explanation of it.

PR #215, opened 2025-01-07
Use containerized database in devcontainer and all unit tests

alexsapps:


mdzirbel (approved):
LGTM. But I'm not that familiar with the code base yet

jakehobbs (approved):
not familiar w/ devcontainers, but seems fine to me

PR #216, opened 2025-01-21
Development setup improvements

alexsapps:


.devcontainer/devcontainer.json
```diff
@@ -42,10 +47,14 @@
 
 	// Use 'forwardPorts' to make a list of ports inside the container available locally.
 	"forwardPorts": [
-		3000, /* Frontend React dev server (adb.dxe.io/v2) */
-		8080, /* ADB main site (adb.dxe.io) */
-		8081 /* Members app (members.dxesf.org) */
-	]
+		8080 /* ADB main site (adb.dxe.io) */
```
jakehobbs, line 50: purposefully removed 8081?

jakehobbs (approved):

.devcontainer/devcontainer.json
alexsapps: yeah because it's noisy when doing run_all and you get two notifications or one at random, and we will almost always want to actually use the 8080 one.  it will still be auto forwarded though, just silently.

PR #218, opened 2025-01-27
Make launch configurations for VS Code and other small improvements.

alexsapps:
This allows the developer to see which programs are still running and view their output separately.


server/debug.env
```diff
@@ -0,0 +1,5 @@
+TEMPLATES_DIRECTORY=../templates
```
jakehobbs, line 1: do we want these both here & in the makefile?

jakehobbs (approved):

PR #229, opened 2025-03-02
[NOT READY, WIP] react attendance page

jakehobbs:


frontend-v2/components.json
```diff
@@ -1,13 +1,13 @@
...
   "style": "new-york",
-  "rsc": false,
+  "rsc": true,
   "tsx": true,
   "tailwind": {
     "config": "tailwind.config.ts",
     "css": "src/styles/globals.css",
     "baseColor": "neutral",
-    "cssVariables": false,
+    "cssVariables": true,
```
jakehobbs, line 10: i don't think we want this actually

frontend-v2/src/lib/utils.ts
```diff
@@ -1,9 +1,6 @@
-import { clsx, type ClassValue } from 'clsx'
-import { twMerge } from 'tailwind-merge'
+import { clsx, type ClassValue } from "clsx"
+import { twMerge } from "tailwind-merge"
 
-// Generated by shadcn. See https://dev.to/ramunarasinga/cn-utility-function-in-shadcn-uiui-3c4k
```
jakehobbs, line 4: shadcn removed this comment, lol

frontend-v2/src/styles/globals.css
```diff
@@ -28,5 +13,66 @@ body {
 @layer base {
   :root {
     --radius: 0.5rem;
+    --background: 0 0% 100%;
```
jakehobbs, line 16: not sure if we actually want all of these css variables here & in tw config. i need to look into shadcn config more.

frontend-v2/pnpm-lock.yaml
jakehobbs, line 1: seems we are using different pnpm versions. i think i'm pinned to an older version that was required for my old job. so i should probably update...

jakehobbs:
closing in favor of #239

PR #230, opened 2025-03-03
Implement navbar in react & remove "pages" router test page

jakehobbs:


frontend-v2/src/components/nav.tsx
```diff
@@ -0,0 +1,148 @@
...
+          <Image src={logo1} alt="DxE" className="w-auto h-auto" priority />
+        </div>
+        <button
+          aria-label="menu"
+          className={cn('navbar-burger burger', {
+            'is-active': isMobileExpanded,
+          })}
+          onClick={() => setMobileExpanded((prev) => !prev)}
+        >
+          <span aria-hidden="true"></span>
```
alexsapps, line 65: could you add a comment what these 3 `<span>`s are for?

frontend-v2/src/components/nav.tsx
```diff
@@ -0,0 +1,148 @@
...
+        if (it === 'admin') return role === 'admin'
+        if (it === 'organizer') return role === 'admin' || role === 'organizer'
+        if (it === 'attendance')
+          return (
+            role === 'admin' || role === 'organizer' || role === 'attendance'
+          )
+        if (it === 'non-sfbay') return role === 'non-sfbay'
+        return false
+      })
+    },
```
alexsapps, line 41: could this be extracted as a named function, perhaps in the top level of the module, since it isn't super navigation or React related? perhaps called getHasAccess.

frontend-v2/src/components/nav.tsx
```diff
@@ -0,0 +1,148 @@
...
+                    }}
+                  >
+                    {dropdown.label}
+                  </a>
+                  {activeDropdown === dropdown.label && (
+                    <div
+                      className={cn('navbar-dropdown !block')}
+                      onClick={() => setActiveDropdown(dropdown.label)}
+                    >
+                      {dropdown.items.map((item) => {
```
alexsapps, line 96: could the dropdown items be extracted to a variable just before the return statement? i think it would make it easier to follow the flow of the remaining JSX  / see the overall outline of the component, and help reduce such deep nesting.

alexsapps (approved):
Awesome! Left some suggestions but I wouldn't mind doing them myself since they're just some mechanical refactors, if you just wanna merge this and lmk if they sound good to you.

frontend-v2/src/components/nav.tsx
jakehobbs: sure i can extract it

frontend-v2/src/components/nav.tsx
jakehobbs: sure, lemme refactor some stuff a bit so we don't have to drill everything so much

frontend-v2/src/components/nav.tsx
jakehobbs: you can write the unit tests :P

PR #234, opened 2025-03-18
Prettier config improvements

alexsapps:
* Move all prettier config to root directory.
* Use prettier on all file types that it supports in VS Code and make fmt for consistency and to avoid spelling out every langauge to use it on.
* Require use of config file to run prettier in VS Code to avoid conflicting defaults between versions of prettier, and add config at root of project.
* Format unformatted files (some mjs files were not formatted previously).


frontend-v2/pnpm-lock.yaml
jakehobbs, line 1: is this getting formatted? i think ideally lock files that are autogenerated don't get touched.

jakehobbs (approved):

frontend-v2/pnpm-lock.yaml
alexsapps: whoops. fixed that and also realized prettier configs can't be nested so i moved all the config to the root directory.

PR #235, opened 2025-03-19
Chapter list 'last action' column color/tooltip

alexsapps:
Updated color ranges based on request from IC (Michelle) and add tooltips for days since last action / days remaining before offboarding.

### DIFF
--- a/frontend/ChapterList.vue
+++ b/frontend/ChapterList.vue
@@ -94,7 +94,6 @@
         <span
           @click="showModal('chapter-last-contact-modal', props.row)"
           class="is-clickable tag"
-          :class="colorQuarterlyGoal(props.row.LastContact)"
           >{{ props.row.LastContact || 'None' }}</span
         >
       </b-table-column>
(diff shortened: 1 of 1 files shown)


frontend/ChapterList.vue
```diff
@@ -754,12 +775,14 @@ export default Vue.extend({
 
       const csrfToken = $('meta[name="csrf-token"]').attr('content');
       this.disableConfirmButton = true;
+      const data = JSON.stringify(this.currentChapter);
+      console.log(data);
```
jakehobbs, line 779: purposefully leaving this logging?

frontend/ChapterList.vue
```diff
@@ -865,19 +888,44 @@ export default Vue.extend({
       }
       return c;
     },
-    colorQuarterlyGoal(text: string) {
+    colorLastAction(text: string) {
+      const GREEN = 'is-success';
+      const YELLOW = 'is-warning';
+      const RED = 'is-danger';
+      const GRAY = 'is-grey';
+      const BLACK = 'is-black';
```
jakehobbs, line 896: nit: could move these to module scope. i would prob do smth like this fwiw:
```ts
const Colors = {
  GREEN: "is-success",
  YELLOW: "is-warning",
  ...
}
```

frontend/ChapterList.vue
```diff
@@ -865,19 +888,44 @@ export default Vue.extend({
...
       }
-      if (time.isAfter(dayjs().add(-58, 'day'))) {
-        c = 'is-warning';
+
+      if (time.isAfter(dayjs().subtract(30 * 2, 'day'))) {
+        return GREEN;
+      } else if (time.isAfter(dayjs().subtract(30 * 3.5, 'day'))) {
+        return YELLOW;
+      } else if (time.isAfter(dayjs().subtract(30 * 4 + 1, 'day'))) {
+        // + 1 becauase "black" means the chapter should be offboarded, but
```
jakehobbs, line 909: nit: typo
```suggestion
        // + 1 because "black" means the chapter should be offboarded, but
```

jakehobbs (approved):

frontend/ChapterList.vue
alexsapps: oops thanks, removed

frontend/ChapterList.vue
alexsapps: done

frontend/ChapterList.vue
alexsapps: done

PR #237, opened 2025-03-19
Add social fields for international organizers

alexsapps:
Looks like this:

image

### DIFF
--- a/frontend/ChapterList.vue
+++ b/frontend/ChapterList.vue
@@ -408,7 +408,12 @@
             <p v-if="!currentChapter.ChapterID">
               Please save the new chapter before adding organizers.
             </p>
-            <b-table :data="currentChapter.Organizers" v-if="currentChapter.ChapterID">
+            <b-table
+              :data="currentChapter.Organizers"
+              detailed
+              :show-detail-icon="true"
+              v-if="currentChapter.ChapterID"
+            >
               <template #empty>
                 <div class="has-text-centered">No organizers found. Add one below.</div>
               </template>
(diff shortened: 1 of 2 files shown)


frontend/ChapterList.vue
```diff
@@ -415,19 +420,51 @@
...
-                  icon="facebook"
-                ></b-input>
-              </b-table-column>
               <b-table-column v-slot="props">
                 <b-button @click="deleteOrganizer(props.row)">
                   <b-icon icon="delete" type="is-danger"></b-icon>
                 </b-button>
               </b-table-column>
+              <template #detail="props">
+                <div style="display: flex; flex-wrap: wrap; gap: 10px">
```
jakehobbs, line 429: nit: i feel like there's a lot of space here. maybe gap-x 10px and gap-y 5px would look a little nicer

jakehobbs: just feels weird to me that there is more vertical space b/w these than there is b/w the "name email phone" fields and the top of this section.

frontend/ChapterList.vue
```diff
@@ -415,19 +420,51 @@
...
+                  <b-field label="Facebook" label-position="on-border" style="max-width: 300px">
+                    <b-input
+                      type="text"
+                      v-model="props.row.Facebook"
+                      placeholder="Facebook"
+                      icon="facebook"
+                      maxlength="100"
+                    ></b-input>
+                  </b-field>
+                  <b-field label="Instagram" label-position="on-border" style="max-width: 300px">
```
jakehobbs, line 439: i wonder if we should do this style of labels for the "name email phone" fields too to keep it a bit more consistent?

server/src/model/chapters.go
```diff
@@ -63,10 +63,13 @@ type ChapterWithToken struct {
...
-	Email    string
-	Phone    string
-	Facebook string
+	Name      string
+	Email     string
+	Phone     string
+	Facebook  string
+	Instagram string
+	Twitter   string
+	Website   string
```
jakehobbs, line 72: when reading organizers out of the db that don't already have these fields set in the json, i assume it will be fine and just return go's nil values (empty string)? my go is rusty....

jakehobbs (approved):

frontend/ChapterList.vue
alexsapps: the table already has a header row that labels the columns themselves so individual labels aren't really necessary in table cells. lmk if i'm missing something though. and max-width doesn't matter as much since the table already forces 3 textboxes to share one line so those elements are never too wide.

server/src/model/chapters.go
alexsapps: yeah i think so. i was able to add fields here after creating test organizers and they loaded fine.

PR #239, opened 2025-04-06
[NOT READY, WIP] react attendance page (take 2)

jakehobbs:


frontend-v2/src/app/event/event-form.tsx
```diff
@@ -0,0 +1,532 @@
...
+import { API_PATH, apiClient } from '@/lib/api'
+import { useQuery, useQueryClient, useMutation } from '@tanstack/react-query'
+import { useParams, useRouter } from 'next/navigation'
+import toast from 'react-hot-toast'
+import { useAuthedPageContext } from '@/hooks/useAuthedPageContext'
+import { SF_BAY_CHAPTER_ID } from '@/lib/constants'
+import { ActivistRegistry } from '@/lib/activist-registry'
+import { AttendeeInputField } from './attendee-input-field'
+
+// TODO(jh):
```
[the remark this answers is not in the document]
jakehobbs: probably should make asana tasks or something for the rest

frontend-v2/src/lib/activist-registry.ts
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    return this.activists
+      .filter(({ name }) => nameFilter(name, input))
+      .slice(0, maxResults)
+      .map((a) => a.name)
+  }
+}
+
+// Like the Vue nameFilter: internal whitespace matches anything.
```
alexsapps: remove this comment? i figure it will be meaningless as soon as the old page is removed.

frontend-v2/src/lib/activist-registry.ts
alexsapps: should this be moved to the app/event directory? this also seems quite specific to event attendance, e.g. exposing hasPhone rather than say a "getPhone" function, and storing only name/email/phone when there are dozens of other possible columns that we probably won't be adding here.

i think it makes sense to keep here if we are going to build on this for the activist list pages too.

frontend-v2/src/app/event/attendee-input-field.tsx
```diff
@@ -0,0 +1,176 @@
...
+        return
+      }
+      case 'Enter': {
+        e.preventDefault()
+        const trimmedValue = field.state.value?.trim() ?? ''
+        if (!trimmedValue.length) {
+          return
+        }
+        const selectedValue =
+          selectedSuggestionIndex >= 0 && suggestions[selectedSuggestionIndex]
```
alexsapps: can suggestions[selectedSuggestionIndex] be falsy when selectedSuggestionIndex >= 0?

frontend-v2/src/app/event/event-form.tsx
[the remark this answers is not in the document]
alexsapps: here you go, can you make separate // TODO lines for each of these?

* cache autofill names:
https://app.asana.com/1/71341131816665/project/1209217418568645/task/1212688232815554

* backup unsaved changes locally:
https://app.asana.com/1/71341131816665/project/1209217418568645/task/1212688232815556

dark mode probably doesn't need a TODO (unless it's for this page only?) but i made a task for it:
https://app.asana.com/1/71341131816665/project/1209217418568645/task/1212688232815552

i'm sure you'll remember to improve styling and test in prod before making it the default in prod?

feel free to keep a // TODO for not returning a redirect. that probably doesn't need an Asana task.

frontend-v2/src/app/event/event-form.tsx
```diff
@@ -0,0 +1,525 @@
...
+        event_date: value.eventDate,
+        event_type: value.eventType,
+        added_attendees: addedAttendees,
+        deleted_attendees: deletedAttendees,
+        suppress_survey: value.suppressSurvey,
+      })
+    },
+  })
+
+  // Helper functions.
```
alexsapps: i don't think this comment adds anything. all of our functions are helpful, hopefully, and it's clear why these functions are here and what they're for.

frontend-v2/src/app/event/event-form.tsx
```diff
@@ -0,0 +1,525 @@
...
+  const checkForDuplicate = (value: string, currentIndex: number): boolean => {
```
alexsapps: consider using a normal function declaration for hoisting and improved readability and debugging

(arrow functions are useful in classes because of `this` binding, but that advantage does not apply in react function components)

frontend-v2/src/app/event/event-form.tsx
```diff
@@ -0,0 +1,525 @@
...
+        matcher.nameContains &&
+        eventName.toLowerCase().includes(matcher.nameContains)
+      ) {
+        return true
+      }
+      // If event type, check for match.
+      if (matcher.type && matcher.type === eventType) {
+        return true
+      }
+      return false
```
alexsapps: if a matcher contained both a nameContains and type, i would expect this to check both before returning true.

```suggestion
      // If event name, check if included (case-insensitive).
      if (
        matcher.nameContains &&
        !eventName.toLowerCase().includes(matcher.nameContains)
      ) {
        return false
      }
      // If event type, check for match.
      if (matcher.type && matcher.type !== eventType) {
        return false
      }
      return true
```

frontend-v2/src/app/event/event-form.tsx
```diff
@@ -0,0 +1,525 @@
...
+            <div className="flex items-center gap-2">
+              <Checkbox
+                id="suppressSurvey"
+                checked={field.state.value}
+                onCheckedChange={(checked) =>
+                  field.handleChange(Boolean(checked))
+                }
+              />
+              <Label htmlFor="suppressSurvey" className="cursor-pointer">
+                Don&apos;t send survey
```
alexsapps: should this be a "Send survey" checkbox that is checked by default when this field is visible?

we could change that later if you just want to replicate the old behavior for now

frontend-v2/src/app/event/event-form.tsx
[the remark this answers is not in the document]
jakehobbs: would like to keep my little list so i remember. i added separate TODOs for the later ones.

frontend-v2/src/lib/activist-registry.ts
jakehobbs: i was thinking maybe the app directory should only contain pages & components, but fine w/ moving it there if you prefer. i don't _think_ it will be used on activist list pages.

frontend-v2/src/app/event/event-form.tsx
```diff
@@ -0,0 +1,525 @@
...
+      eventType: eventData?.event_type || (isConnection ? 'Connection' : ''),
+      eventDate: eventData?.event_date || '',
+      // For new events, non-SF Bay chapters default to not sending surveys.
+      suppressSurvey:
+        eventData?.suppress_survey ?? user.ChapterID !== SF_BAY_CHAPTER_ID,
+      attendees:
+        eventData?.attendees && eventData.attendees.length > 0
+          ? [
+              ...eventData.attendees.map((name) => ({ name })),
+              ...Array(Math.max(0, MIN_EMPTY_FIELDS))
```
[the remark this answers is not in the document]
jakehobbs: 💀

frontend-v2/src/app/event/event-form.tsx
jakehobbs: makes sense

frontend-v2/src/app/event/event-form.tsx
jakehobbs: slight preference to leave as-is for now & address during UX/UI improvement phase. will add a TODO.

frontend-v2/src/app/event/event-form.tsx
```diff
@@ -0,0 +1,525 @@
...
+    if (emptyCount < MIN_EMPTY_FIELDS) {
+      form.pushFieldValue('attendees', { name: '' })
+    }
+  }
+
+  // Warn before leaving with unsaved changes.
+  useEffect(() => {
+    const handleBeforeUnload = (e: BeforeUnloadEvent) => {
+      if (form.state.isDirty) {
+        e.preventDefault()
```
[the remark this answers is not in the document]
jakehobbs: The warning is handled by the browser itself. When you call e.preventDefault() and set e.returnValue = '' in a beforeunload event handler, the browser automatically shows its own built-in dialog asking the user if they really want to leave the page.

The code doesn't create a custom warning message - it just triggers the browser's native "Leave site? Changes you made may not be saved" dialog (the exact wording varies by browser). Modern browsers ignore any custom message you try to set in e.returnValue for security reasons, so the empty string '' is just a convention to ensure the event is properly triggered across different browsers.

adding comment

frontend-v2/src/app/event/activist-registry.ts
```diff
@@ -0,0 +1,53 @@
...
+  private names: Set<string>
+  private activists: ActivistRecord[]
+
+  constructor(activists: ActivistRecord[]) {
+    this.activists = activists
+    this.byName = new Map(activists.map((a) => [a.name, a]))
+    this.names = new Set(activists.map((a) => a.name))
+  }
+
+  hasEmail(name: string): boolean {
```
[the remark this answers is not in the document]
jakehobbs: not concerned since it's constant time, but i'll combine the email & phone lookup.

frontend-v2/src/app/event/event-form.tsx
jakehobbs: every react codebase i've ever worked in almost always uses arrow functions within react components. they are more succinct & readable imo. chatgpt agrees it's standard:

Screenshot 2026-01-10 at 8 32 35 PM

Screenshot 2026-01-10 at 8 32 50 PM

Screenshot 2026-01-10 at 8 33 04 PM

frontend-v2/src/app/event/page.tsx
alexsapps: should this file be moved to event/new/page.tsx, to make room for a future event/page.tsx that lists all the events?

and should /event be renamed to /events for consistency with /users? cursory google search says plural is preferred

sorry just noticed this now

alexsapps (approved):

frontend-v2/src/app/event/page.tsx
jakehobbs: i think i was thinking that `/event/list` would be for the event list. no strong feelings, but will revisit later.

alexsapps: ok sounds good to revisit later. just for the record my reasoning was it's easier to let users change the url themselves, at least to go up, e.g. if they see `/events/218` or `/events/add` they can remove `/218` or `/add` to go up a level in the navigation. i think of the list page as the highest level in that navigation hierarchy because it will usually provide all the other links.

alexsapps:
i just tried this out in prod, and i noticed some changes don't seem to save.

to reproduce:
1. open an existing event
2. remove someone, hit save
3. add same person back, hit save
result: their name disappears again / they do not get added back

similarly:
1. open an existing event
2. add someone, hit save
3. remove the person you just added, hit save
result: they do not get removed / their name reappears

PR #241, opened 2025-04-06
Undo navigation bug on event list page

alexsapps:
It was showing the navigation as if the user was always an admin. Introduced here, probably for testing: https://github.com/dxe/adb/commit/f02c797c0224d0076e84ebc2ebe6a4d68c2ef104#diff-cb8123a015cec42248e9e3799392ad0185efe8c5b725d024ca068afa333181c5L4

### DIFF
--- a/server/templates/event_list.html
+++ b/server/templates/event_list.html
@@ -1,12 +1,7 @@
 {{template "header.html" .}}
 
 <div id="app">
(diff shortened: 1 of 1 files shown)


jakehobbs (approved):
Thanks, I think ChatGPT did this and I didn't catch it.

PR #242, opened 2025-04-07
Replace 47 with SFBayChapterId variable

alexsapps:


server/src/form_processor/queries.go
```diff
@@ -189,7 +191,7 @@ select
         '',
         '', 
         NULL,
-        '47',
+        '` + model.SFBayChapterIdStr + `',
```
jakehobbs, line 194: ❓  does this not work for this one?

```suggestion
        model.SFBayChapterIdStr,
```

jakehobbs (approved):
def nicer than what we had before, but curious your thoughts on handling this long-term. do you think sf bay will always be a "special case" and/or could we handle this more dynamically?

server/src/form_processor/queries.go
alexsapps: this is a list in SQL in the middle of a huge `backtick string`, not a Golang function call or other list in Go :)

alexsapps:
> def nicer than what we had before, but curious your thoughts on handling this long-term. do you think sf bay will always be a "special case" and/or could we handle this more dynamically?

I think it can make sense to try things out on the main chapter or greater bay area generally before applying them to other chapters. We could have dynamic "chapter settings" thing and add a new setting when we want to experiment on a chapter level, but if it's always just main chapter vs all chapters, then that might not be necessary. We could just wait until it gets annoying and then change it later / make use of it incrementally as needed. I'm not seeing a situation where we would get locked into something that's really hard to change. Perhaps it would be good to have a rule that we don't hard-code more than 3 chapter IDs for one feature -- once we want to add a 4th then we know it's a real thing and we make a chapter settings mechanism for it.

PR #243, opened 2025-04-23
Revamp international application emails

alexsapps:
Add several new types of emails and rework the package's architecture. See new README for details.


server/src/international_application_processor/mailer_common.go
```diff
@@ -0,0 +1,99 @@
...
+
+func validateEmail(str string) error {
+	_, err := mail.ParseAddress(str)
+	return err
+}
+
+func sanitizeState(state string) string {
+	return strings.Map(func(r rune) rune {
+		if unicode.IsLetter(r) {
+			return r
```
jakehobbs, line 95: should this also make it uppercase if it's not or nah?

server/src/international_application_processor/onboarding_messages.go
```diff
@@ -0,0 +1,327 @@
...
+	fmt.Fprintf(&body, "<p>Hey %v!</p>", html.EscapeString(b.firstName))
+
+	body.WriteString(
+		`<p>
+			Thank you for reaching out about getting involved with Direct Action
+			Everywhere.
+		</p>
+		<p>
+			Great news—the SF Bay Area chapter is within 100 miles of you.
+			We’d love to see you at an upcoming event!
```
jakehobbs, line 40: nit: this sounds kinda robotic ("within 100 miles of you")... should we say something a bit more conversational like "your near the SF Bay chapter" or something?

server/src/international_application_processor/onboarding_messages.go
```diff
@@ -0,0 +1,327 @@
...
+	fmt.Fprintf(&body, `
+		<p>
+			To begin, please find 1-2 other people in your area who might be
+			interested in helping you organize a chapter. Once you have
+			identified them, reach out to me by email at %v to schedule a call.
+			Don’t hesitate to email me with any questions in the meantime.
+			I’m looking forward to hearing back from you,
+		</p>`, californiaCoordinator.Address)
+
+	body.WriteString(almiraTannerSignatureHtml)
```
jakehobbs, line 186: nit: this code feels like a little bit of a footgun if we update the `californiaCoordinator` to be someone else, but forget that it includes `almiraTannerSignatureHtml`? i wonder if it's worth adding `role` to the coordinator structs & then just building the signature for everyone based on that.

server/src/international_application_processor/onboarding_messages.go
```diff
@@ -0,0 +1,327 @@
...
+			Thank you for your interest in joining the DxE Network.
+			We recently launched the Network Member Program which is a platform
+			for people who are interested in taking action with DxE but don’t
+			have a chapter near them or have the capacity to organize a chapter
+			themselves.
+			The only steps to complete to officially become a member are:
+		</p>
+		<ol>
+			<li>Watch a video</li>
+			<li>Complete a short quiz</li>
```
jakehobbs, line 268: ❓ are these supposed to be links?

jakehobbs (approved):
looks amazing!

server/src/international_application_processor/onboarding_messages.go
alexsapps: oh good idea. i felt this was a footgun but it didn't occur to me to just add a "role" field. i moved the roles for Michelle and Almira and made individual functions for each coordinator role, but i didn't make a single generic signature function because they're not all exactly the same. i don't want to make everyone standardize on whether to put their pronouns, and it's nice to allow small customizations, like Antonelle's sig says "DxE Organizer" instead of "Organizer\nDxE".

anyway hopefully we'll catch errors with manual review of the HTML allowed by mailer_integration_test.go.

server/src/international_application_processor/onboarding_messages.go
alexsapps: yes, waiting for Michelle. will do something here before merging.

alexsapps: done

PR #248, opened 2025-06-25
Combine intl app onboarding/notification emails

alexsapps:


alexsapps:
image
Emails look like this now

jakehobbs (approved):
seems a little strange imo to combine them, but maybe there's context behind the decision that i'm unaware of.

code lgtm!

alexsapps:
> seems a little strange imo to combine them, but maybe there's context behind the decision that i'm unaware of.
> 
> code lgtm!

Thanks!  I explained in the commit description, but I updated that after creating the PR and forgot to sync the PR description with it. Duplicating it here for reference:

"The notification emails were supposed to [go to] the same people as the
onboarding emails, but there was the potential for this to get out of
sync, and it did. This simplifies the code while also sending fewer
emails to organizers, and it allows the responders to see their own
responses for reference.

This also splits `mailer_common.go` into `chapter_emails.go` and
`form_validation.go`."

PR #261, opened 2025-07-07
Fix issues from chapter switcher

alexsapps:
fix running vue frontend in development
fix an issue that limited the columns SF Bay chapter users could view


alexsapps:
merging before review since Antonelle is waiting for this :)

PR #271, opened 2025-07-25
feat: provide chapter id to signup service and add endpoint to get chapter details

alexsapps:


server/src/main.go
```diff
@@ -262,7 +262,10 @@ func router() (*mux.Router, *sqlx.DB) {
 	router.Handle("/health", alice.New(main.corsAllowGetMiddleware).ThenFunc(main.HealthStatusHandler))
 	router.Handle("/static_resources_hash", alice.New(main.corsAllowGetMiddleware).ThenFunc(main.StaticResourcesHashHandler))
 	router.Handle("/external_events/{page_id:[0-9]+}", alice.New(main.corsAllowGetMiddleware).ThenFunc(main.ListFBEventsHandler))
-	router.Handle("/chapters/{lat:[0-9.\\-]+},{lng:[0-9.\\-]+}", alice.New(main.corsAllowGetMiddleware).ThenFunc(main.FindNearestChaptersHandler))
+	// Deprecated. Use "/chapters/nearest/{lat:[0-9.\\-]+},{lng:[0-9.\\-]+}" instead, which returns real Chapter ID in response `id` field rather than Facebook Page ID.
```
jakehobbs, line 265: should we include a comment about why this is being marked as deprecated vs just being removed, and if/when there are plans to remove it?

jakehobbs (approved):

server/src/main.go
jakehobbs: i assume we are only keeping it b/c the public website is using it. once we update that, i think we're good to remove it?

PR #274, opened 2025-08-10
feat: add timestamp columns to activists

alexsapps:
* feat: add timestamp columns to activists
This allows smarter merging of activist records. This is especially
important because in the future we may decide to merge activist records
across chapters. We suspect there may be a lot of activists in multiple
chapters now that check-in forms systematically duplicate activists
across chapters when someone gives a zip code near one chapter (which is
presumed to be their home chapter) on a check-in form belonging to
another chapter (which needs the same activist info for attendance).

* chore: deduplicate update queries for activist update
One was missing dev_application_date and dev_application_type and the
other was missing lat, lng and coords_updated. Combining so all fields
will get updated in all cases.

### DIFF
--- a/server/scripts/db-migrations/20250810034041_activist_timestamps.up.sql
+++ b/server/scripts/db-migrations/20250810034041_activist_timestamps.up.sql
@@ -1,0 +1,22 @@
+-- *_updated timestamps may be used to resolve conflicts when merging activist records
+
+-- A sentinel value is used as the default in order to indicate unknown modification times for
+-- existing table rows in production.
+ALTER TABLE activists
+    ADD COLUMN created TIMESTAMP NOT NULL DEFAULT '1970-01-01 00:00:01',
+    ADD COLUMN phone_updated TIMESTAMP NOT NULL DEFAULT '1970-01-01 00:00:01',
+    ADD COLUMN email_updated TIMESTAMP NOT NULL DEFAULT '1970-01-01 00:00:01',
+    ADD COLUMN address_updated TIMESTAMP NOT NULL DEFAULT '1970-01-01 00:00:01',
+    ADD COLUMN coords_updated TIMESTAMP NOT NULL DEFAULT '1970-01-01 00:00:01';
+
+-- Change default for new records to CURRENT_TIMESTAMP.
+ALTER TABLE activists
+    MODIFY COLUMN created TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP;
+ALTER TABLE activists
+    MODIFY COLUMN phone_updated TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP;
+ALTER TABLE activists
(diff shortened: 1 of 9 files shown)


server/scripts/db-migrations/20250810034041_activist_timestamps.up.sql
```diff
@@ -0,0 +1,22 @@
+-- *_updated timestamps may be used to resolve conflicts when merging activist records
+
+-- A sentinel value is used as the default in order to indicate unknown modification times for
+-- existing table rows in production.
+ALTER TABLE activists
+    ADD COLUMN created TIMESTAMP NOT NULL DEFAULT '1999-12-31 00:00:00',
```
jakehobbs, line 6: lol i'm surprised the default isn't 1970-01-01.

server/scripts/db-migrations/20250810034041_activist_timestamps.up.sql
```diff
@@ -0,0 +1,22 @@
...
+    ADD COLUMN phone_updated TIMESTAMP NOT NULL DEFAULT '1999-12-31 00:00:00',
+    ADD COLUMN email_updated TIMESTAMP NOT NULL DEFAULT '1999-12-31 00:00:00',
+    ADD COLUMN address_updated TIMESTAMP NOT NULL DEFAULT '1999-12-31 00:00:00',
+    ADD COLUMN coords_updated TIMESTAMP NOT NULL DEFAULT '1999-12-31 00:00:00';
+
+-- Change default for new records to CURRENT_TIMESTAMP.
+ALTER TABLE activists
+    MODIFY COLUMN created TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP;
+ALTER TABLE activists
```
jakehobbs, line 15: hmmm, does this mean that when an activist is created, the `phone_updated` would be set to the current timestamp even if they have no phone number? from a db perspective, i feel like having these nullable makes the most sense. but if you've already thought about this, i'm fine w/ it.

server/src/model/activist.go
```diff
@@ -272,10 +273,13 @@ SET
   street_address = :street_address,
   city = :city,
   state = :state,
+  address_updated = IF(street_address <> :street_address OR city <> :city OR state <> :state, NOW(), address_updated),
+  lat = :lat,
+  lng = :lng,
```
jakehobbs, line 278: i think the coords were purposefully not here before, b/c they are fields that are only set (and i believe only consumed) via the backend and are not exposed to users of the adb.

but i suppose this is fine, as long as the existing values are actually making it out to the frontend.

jakehobbs (approved):
could of questions, but lgtm

server/src/model/activist.go
jakehobbs: looks like it's in the `ActivistJSON` struct, so should be fine.

alexsapps: looks like i missed buildActivistJSONArray and CleanActivistData. thanks for helping me catch this!

jakehobbs (approved):

jakehobbs (dismissed):

PR #277, opened 2025-08-15
feat: new list view: supporters pending workshop

RunicTable3626:
View for all 'new' activists that have not yet done a workshop

I think new should be defined as: 	
-someone who is labeled as a "supporter" under the column level 
-has attended their very first dxe event in the last 6 months 
-has not taken the workshop yet 
-and no limit to how many events they've attended in the last 6 months

Also addition requests for fields by default by Antonelle:

The visible columns that would be great to have by default are:

-name
-email
-phone number
-level
-zip code
-total events
-last event
-mpi
-workshop

And having an option to export csv

### DIFF
--- a/server/src/main.go
+++ b/server/src/main.go
@@ -633,7 +633,7 @@ func (c MainController) NewActivistsPendingWorkshopHandler(w http.ResponseWriter
 		PageName: "NewActivistsPendingWorkshopList",
 		Data: ActivistListData{
 			Title:       "New Activists Pending Workshop",
-			Description: "Everyone who has their first event within the given range (last 6 months by default) AND has not yet attended the Intro Workshop.",
+			Description: "Supporters whose first event was within the given range (last 6 months by default) AND have not yet attended the Intro Workshop.",
 			View:        "new_activists_pending_workshop",
 		},
 	})


server/src/main.go
```diff
@@ -240,6 +240,7 @@ func router() (*mux.Router, *sqlx.DB) {
 	router.Handle("/list_connections", alice.New(main.authOrganizerMiddleware).ThenFunc(main.ListConnectionsHandler))
 	router.Handle("/list_activists", alice.New(main.authOrganizerOrNonSFBayMiddleware).ThenFunc(main.ListActivistsHandler))
 	router.Handle("/new_activists", alice.New(main.authOrganizerOrNonSFBayMiddleware).ThenFunc(main.NewActivistsHandler))
+	router.Handle("/new_activists_pending_workshop", alice.New(main.authOrganizerMiddleware).ThenFunc(main.NewActivistsPendingWorkshopHandler))
```
alexsapps, line 243: were you intending this to be for SF bay only or all chapters?

if all chapters, i think you either need authOrganizerOrNonSFBayMiddleware here and apiOrganizerOrNonSFBayAuthMiddleware below, otherwise need to add "roleRequired": ["organizer"] in shared/nav.json to make it appear only in the SF bay chapter (since "organizer" role currently effectively means "sf bay organizer")

server/src/main.go
```diff
@@ -626,6 +628,17 @@ func (c MainController) NewActivistsHandler(w http.ResponseWriter, r *http.Reque
 	})
 }
 
+func (c MainController) NewActivistsPendingWorkshopHandler(w http.ResponseWriter, r *http.Request) {
+	renderPage(w, r, "activist_list", PageData{
+		PageName: "NewActivistsPendingWorkshopList",
+		Data: ActivistListData{
+			Title:       "New Activists Pending Workshop",
+			Description: "Everyone who has their first event within the given range (last 6 months by default) AND has not yet attended the workshop.",
```
alexsapps, line 636: i wouldn't say "workshop" is so widely understood to mean the intro workshop. actually i didn't even know we had something called that unless it's the how-to-change-the-world-in-one-generation workshop? anway we naturally may have lots of other kinds of workshops.

```suggestion
			Description: "Everyone who has their first event within the given range (last 6 months by default) AND has not yet attended the Intro Workshop.",
```

alexsapps (approved):

server/src/model/activist.go
```diff
@@ -861,6 +861,18 @@ func GetActivistsExtra(db *sqlx.DB, options GetActivistOptions) ([]ActivistExtra
 				) BETWEEN ? AND ?`)
 			queryArgs = append(queryArgs, options.LastEventDateFrom, options.LastEventDateTo)
 
+		case "new_activists_pending_workshop":
+			whereClause = append(whereClause, `
```
alexsapps, line 865: i just noticed this in the description as i was about to merge:
'-someone who is labeled as a "supporter" under the column level'

i'm not actually sure what Antonelle wants but do you know or want to check if we should filter on this? i learned since last time that "supporter" not only distinguishes activists from "organizer"s, but also from "non-local"s which we never set automatically but ADB users manually set people to be non-local.

if not then we can just remove this from the PR description

alexsapps (changes requested):

server/src/model/activist.go
RunicTable3626: so "supporter" only refers to supported in the SF Bay Area? I'll reach out to Antonelle and see what she thinks about this.

alexsapps (changes requested):
Sorry, one last thing!

alexsapps (approved):

PR #279, opened 2025-08-30
Auth and logging fixes

alexsapps:
* refactor: use constants for SF Bay ids

* test: use real database and cookies for dev test user

This removes a noisy message about not getting the chapter ID from the
cookie when there was no cookie.

* chore: replace fmt.Print* with log.Print* statements

### DIFF
--- a/server/src/main.go
+++ b/server/src/main.go
@@ -84,7 +84,6 @@ func augmentUserWithChapterFromSession(db *sqlx.DB, r *http.Request, adbUser mod
 
 	chapterID, ok := authSession.Values["chapterid"].(int)
 	if !ok {
-		// User could have old session cookie that doesn't contain `chapterid`.
 		return adbUser, fmt.Errorf("failed to get chapter ID from session")
 	}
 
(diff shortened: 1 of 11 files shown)


server/src/main.go
```diff
@@ -98,18 +97,25 @@ func augmentUserWithChapterFromSession(db *sqlx.DB, r *http.Request, adbUser mod
...
+		adbUser, authed = getAuthedADBUser(db, r)
 	}
 
+	return adbUser, authed
+}
+
+func getAuthedADBUser(db *sqlx.DB, r *http.Request) (adbUser model.ADBUser, authed bool) {
 	// First, check the cookie.
-	authSession, err := sessionStore.New(r, "auth-session")
+	authSession, err := sessionStore.Get(r, "auth-session")
```
jakehobbs, line 118: do we not need to call `.New` anywhere?

jakehobbs, line 118: do we not need to call `.New` anywhere?

jakehobbs (approved):

PR #280, opened 2025-08-31
fix: avoid overlapping form processor runs

alexsapps:
The form processor was causing race conditions with itself, when two instances would go to insert the same activist. This was causing noisy alerts in AWS.

### DIFF
--- a/server/src/config/config.go
+++ b/server/src/config/config.go
@@ -82,11 +82,6 @@ var (
 		"@every 10s",
 		false,
 	)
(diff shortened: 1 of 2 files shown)


jakehobbs:
Interesting, I thought we only ran the form processor on the main instance.

alexsapps:
> Interesting, I thought we only ran the form processor on the main instance.

We do, but every 10 seconds, so it would still overlap within the same instance without SkipIfStillRunning

jakehobbs:
Oh i didn't realize it ran that often now. 👌

PR #395, opened 2026-05-14
fix: don't ignore auth save error

alexsapps:
## Bug Fixes
- Authentication sessions now handle save failures properly, ensuring errors are surfaced instead of ignored.
- Token-based sign-in stops and returns an error if session saving fails, preventing invalid session states.
- Improved logging and error responses when updating active session state to increase reliability of authentication.

[Review Change Stack](https://app.coderabbit.ai/change-stack/dxe/adb/pull/395)


alexsapps:
@coderabbitai review

PR #429, opened 2026-05-31
feat(events): support scheduling public events in advance

jakehobbs:
Adds a richer event model so events can be created ahead of time — start/end time, IANA timezone, description, online flag, and public flag — surfaced in the event form via a scheduled-event flow (Places address autocomplete + timezone-aware time fields) that appears when "Public event" is checked. Location is stored denormalized on the event itself (a free-text display name plus optional Google Place id and coordinates) rather than in a shared table. After creating a public event the user now lands on a confirmation page summarizing the event with next-step choices (take attendance, edit, create another, done) instead of being dropped on the attendance page. Also introduces a Home hub listing today's events, repoints the post-login redirect and nav logo to `/home`, and serves a referrer-restricted Google Places key through the authed user-info endpoint. Backend changes include the migration for the new event columns, event model read/write of the new fields, and `wipe`/seed/Makefile housekeeping.

## Release Notes

* **New Features**
  * Added event confirmation interface with event details and navigation options.
  * Introduced home dashboard displaying today's events with real-time "Happening now" status.
  * Enabled advanced event scheduling with date, time, and timezone support.
  * Added location selection for events using Google Places integration.
  * Implemented collapsible event details summary for easier event editing.
  * Updated event creation workflow with improved form organization and field gating.

* **Documentation**
  * Updated setup instructions for Google Places API key configuration.


frontend-v2/src/app/(authed)/events/places-autocomplete.tsx
```diff
@@ -0,0 +1,198 @@
...
+          status === 'error'
+            ? 'Location search unavailable'
+            : 'Search for a place'
+        }
+        onChange={(e) => {
+          setText(e.target.value)
+          // Clearing the text clears the selected place. Typing without
+          // selecting a suggestion does not submit a free-text location.
+          if (e.target.value.trim() === '') onClear()
+        }}
```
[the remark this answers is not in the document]
jakehobbs: Fixed. The input now clears the committed place on any manual edit (guarded by a programmaticEditRef so a dropdown selection or parent-driven value sync is not mistaken for typing), so the displayed text and the submitted location can no longer drift apart. Same fix as CodeRabbit's comment below.

frontend-v2/src/app/(authed)/home/home-hub.tsx
```diff
@@ -0,0 +1,210 @@
...
+  todayInTimezone,
+} from '@/lib/timezone'
+
+export function HomeHub() {
+  const { user } = useAuthedPageContext()
+
+  // "Today" is computed in the viewer's timezone (a defined zone) rather than
+  // the DB/browser raw date, to avoid off-by-one-day errors. Chapters have no
+  // timezone of their own, so the creator's zone is the best available default.
+  const today = useMemo(() => todayInTimezone(getBrowserTimezone()), [])
```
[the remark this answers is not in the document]
jakehobbs: Fixed. `today` is now derived from the per-minute `now` clock state (`todayInTimezone(getBrowserTimezone(), now)`), so a tab left open past local midnight rolls over automatically. `todayInTimezone` takes an injectable `now` for this.

server/src/model/event.go
```diff
@@ -512,9 +604,115 @@ func CleanEventData(db *sqlx.DB, body io.Reader, chapterID int) (Event, error) {
...
+	}
+
+	_, err = tx.Exec(`INSERT INTO locations (chapter_id, google_place_id, name, formatted_address, lat, lng)
+VALUES (?, ?, ?, ?, ?, ?)
+ON DUPLICATE KEY UPDATE name = VALUES(name), formatted_address = VALUES(formatted_address), lat = VALUES(lat), lng = VALUES(lng)`,
+		chapterID, placeID, name, formattedAddress, lat, lng)
+	if err != nil {
+		_ = tx.Rollback()
+		return 0, fmt.Errorf("failed to upsert location %s: %w", placeID, err)
+	}
```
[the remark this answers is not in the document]
jakehobbs: Fixed. The upsert no longer clobbers stored metadata with blanks: `name`/`formatted_address` only overwrite when the new value is non-empty (`IF(VALUES(x) <> '  ', ...)`) and `lat`/`lng` use `COALESCE(VALUES(x), x)`, so a later event picking the same place with a sparser Google response keeps the richer record.

frontend-v2/src/app/(authed)/events/event-form.tsx
```diff
@@ -203,7 +320,15 @@ export const EventForm = ({ mode }: EventFormProps) => {
...
+      // Public events carry the richer scheduled-event payload (time, location,
+      // description); quick attendance entry and coaching stay the plain payload.
+      // Pre-existing events with this data keep it editable either way.
+      const includeUpcoming =
+        !isConnection && (value.isPublic || editingHasUpcomingData)
+
+      // Attendees are required for quick attendance entry and coaching, but not
+      // for public events (those are usually filled in later).
+      if (!includeUpcoming && attendeeNames.length === 0) {
         toast.error('At least one attendee is required')
```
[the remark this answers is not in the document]
jakehobbs: This is intentional, not a regression. `editingHasUpcomingData` is latched on purpose: basing `includeUpcoming` on `value.isPublic` alone would drop the scheduled fields (location/time/description) from the save payload when the box is unchecked and silently wipe them. Scheduled events legitimately have zero attendees (attendance is recorded later, at the event), so not requiring an attendee for them is the desired behavior. Quick-attendance events never have upcoming data, so they still require at least one attendee.

frontend-v2/src/app/(authed)/events/[id]/page.tsx
```diff
@@ -35,8 +38,8 @@ export default async function EditEventPage({
   return (
     <ContentWrapper size="sm" className="gap-8">
       <HydrationBoundary state={dehydrate(queryClient)}>
-        <h1 className="text-3xl font-bold">Attendance</h1>
-        <EventForm mode="event" />
+        <h1 className="text-3xl font-bold">Event</h1>
+        <EventForm mode="event" startExpanded={edit === '1'} />
```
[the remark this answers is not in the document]
Comment (jakehobbs) [reply frontend-v2/src/app/(authed)/events/[id]/page.tsx]: Not an issue — this route still serves attendance. `EventForm` is the attendance UI: for an existing event the detail fields collapse behind the summary bar by default (`detailsExpanded = !eventId`) while the attendee section renders immediately (`showAttendeeSection` is only false for a brand-new public event). So "Take attendance now" → `/events/${id}` lands directly on the attendee list, with event details tucked behind the collapsible header. Only the heading changed (Attendance → Event); the flow is intact. No change made.

frontend-v2/src/app/(authed)/events/event-form.tsx
```diff
@@ -23,9 +24,16 @@ import { useAuthedPageContext } from '@/hooks/useAuthedPageContext'
...
 import { DatePicker } from '@/components/ui/date-picker'
 import { format, parseISO } from 'date-fns'
-import { Save } from 'lucide-react'
+import { Save, ChevronDown, ChevronUp } from 'lucide-react'
+import { TimeField } from '@/components/ui/time-field'
+import {
+  getBrowserTimezone,
+  getCommonTimezones,
+  getZoneAbbreviation,
+} from '@/lib/timezone'
```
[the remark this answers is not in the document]
jakehobbs: Fixed. `setDateToToday` now uses `todayInTimezone(form.state.values.timezone || browserTz)` instead of the browser date, so "Today" lands on the correct calendar day in the event's selected timezone around midnight.

frontend-v2/src/app/(authed)/events/places-autocomplete.tsx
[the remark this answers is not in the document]
jakehobbs: Fixed. `onChange` now clears the committed place on any manual edit (not just when emptied), guarded by a `programmaticEditRef` so a dropdown selection or value sync is not treated as user typing. The visible text and submitted location can no longer diverge.

frontend-v2/src/app/(authed)/home/home-hub.tsx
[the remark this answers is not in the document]
jakehobbs: Fixed. `today` is now recomputed from the existing per-minute `now` clock state, so the page rolls over at local midnight without a manual reload.

frontend-v2/src/app/(authed)/home/home-hub.tsx
```diff
@@ -0,0 +1,210 @@
...
+    [today],
+  )
+
+  // Upcoming = today through a year out. This reuses the existing events list
+  // page via its URL date filter rather than duplicating a list here. The
+  // events list keeps its own default (last month → today) when opened directly.
+  const upcomingHref = useMemo(() => {
+    const end = `${Number(today.slice(0, 4)) + 1}${today.slice(4)}`
+    return `/events?start=${today}&end=${end}`
+  }, [today])
```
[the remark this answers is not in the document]
jakehobbs: Fixed. `upcomingHref` now adds a year via `Date` math (`setFullYear(... + 1)`) instead of string slicing, so leap days like 2028-02-29 no longer produce an invalid 2029-02-29.

frontend-v2/src/lib/timezone.ts
```diff
@@ -0,0 +1,143 @@
...
+// Today's calendar date (YYYY-MM-DD) in the given timezone. Used to compute
+// "today's events" in a defined zone rather than the DB/browser raw date.
+export function todayInTimezone(timezone: string): string {
+  // en-CA formats as YYYY-MM-DD.
+  return new Intl.DateTimeFormat('en-CA', {
+    timeZone: timezone,
+    year: 'numeric',
+    month: '2-digit',
+    day: '2-digit',
+  }).format(new Date())
```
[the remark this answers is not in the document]
jakehobbs: Fixed. `todayInTimezone` and `nowMinutesInTimezone` now wrap their `Intl.DateTimeFormat` calls in try/catch and fall back to the local zone, so a bad stored `event.timezone` makes `isEventHappeningNow` return false rather than throwing a RangeError that crashes the whole list render.

pkg/shared/db-migrations/20260530120000_advance_events_and_locations.up.sql
```diff
@@ -0,0 +1,25 @@
...
+-- empty/null so existing attendance rows remain valid.
+ALTER TABLE events
+  ADD COLUMN location_id INT NULL,
+  ADD COLUMN is_online   TINYINT(1) NOT NULL DEFAULT 0,
+  ADD COLUMN description TEXT,
+  ADD COLUMN start_time  TIME NULL,
+  ADD COLUMN end_time    TIME NULL,
+  ADD COLUMN timezone    VARCHAR(64) NOT NULL DEFAULT '',
+  ADD COLUMN is_public   TINYINT(1) NOT NULL DEFAULT 0,
+  ADD CONSTRAINT fk_events_location FOREIGN KEY (location_id) REFERENCES locations (id);
```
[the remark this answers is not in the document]
jakehobbs: Fixed. The migration now adds `UNIQUE KEY uniq_chapter_location (chapter_id, id)` to `locations` and makes the FK composite: `FOREIGN KEY (chapter_id, location_id) REFERENCES locations (chapter_id, id)`. An event can now only reference a location in its own chapter. When `location_id` is NULL the FK is not enforced (MATCH SIMPLE), so existing attendance rows stay valid. Verified the migration applies cleanly in the test-container suite (events.chapter_id and locations.chapter_id are both INT, so no FK type mismatch).

server/src/model/event.go
```diff
@@ -512,9 +604,115 @@ func CleanEventData(db *sqlx.DB, body io.Reader, chapterID int) (Event, error) {
...
+				strings.TrimSpace(eventJSON.Location.FormattedAddress),
+				eventJSON.Location.Lat,
+				eventJSON.Location.Lng,
+			)
+			if err != nil {
+				return Event{}, err
+			}
+			e.LocationID = &locationID
+		}
+	}
```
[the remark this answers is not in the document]
jakehobbs: Partially addressed. I added server-side guards in `CleanEventData`: a public event is now rejected if it lacks a `start_time` or `timezone`, and an in-person public event is rejected without a location. These only trigger when `is_public=true` (the new flow), so connection/quick-attendance/legacy saves are unaffected. I intentionally did NOT add the `end_time` < `start_time` rejection: events that cross midnight (overnight vigils) are a known gap we are deferring (see the TODO on the form's end-time refine), and rejecting them server-side would block adding that support later.

frontend-v2/src/app/(authed)/home/home-hub.tsx
alexsapps, line 1: since this is only the homepage for people with attendance access, and it's only related to events, how about a path like /events/home or /events/today ?

i also feel like the "Welcome" + name/chapter/date is just more boilerplate on a page that already has boilerplate (today's events being the content, and the "Today's events" title and new event / view future events links being boilerplate).

alexsapps, line 1: thoughts on adding this page to the attendance section of the navigation? since some people not might know that clicking on the logo brings them to this page, or might not think of it as the homepage, or anyway might just click the attendance menu out of habit?

server/src/main.go
```diff
@@ -1775,6 +1775,10 @@ func (c MainController) AuthedUserInfoHandler(w http.ResponseWriter, r *http.Req
 
 	writeJSON(w, map[string]interface{}{
 		"user": user,
+		// Referrer-restricted, client-side Google Places key. Served from the
+		// backend config (same source the Vue app uses) so it works in all
+		// environments without baking it into the Next.js build.
```
alexsapps: remove comment - these details not relevant here.
or move to where it's defined

frontend-v2/src/app/authed-page-provider.tsx
```diff
@@ -5,6 +5,8 @@ import { type AuthedUser } from '@/lib/api'
 
 export type TAuthedPageContext = {
   user: AuthedUser
+  // Referrer-restricted Google Places key, served from the backend config.
```
alexsapps: remove comment - not relevant here

Makefile
```diff
@@ -40,7 +40,10 @@ run_all:
 run:
 	cd server/src && go install # Install first so that we keep cached build objects around.
 
-	set -a && . server/debug.env && set +a && \
+	set -a && \
+	. server/debug.env && \
+	{ [ -f .env ] && . ./.env || true; } && \
```
alexsapps: should launch.json be updated too?  🙏:)

alexsapps: can you say how to populate .env in the README, or make a template env file and just mention it in the README? probably say this somewhere: "This API key can be found in GCP under the dev environment."

frontend-v2/src/components/nav.tsx
```diff
@@ -282,9 +282,13 @@ export const Navbar = () => {
       }
     >
       <div className={buefyStyles['navbar-brand']}>
-        <div className={buefyStyles['navbar-item']}>
+        <Link
```
alexsapps, line 285: should we gate this change behind feature flag until we are ready to actually list the events publicly and support RSVP and stuff?

frontend-v2/src/app/(authed)/events/event-form.tsx
```diff
@@ -349,161 +516,422 @@ export const EventForm = ({ mode }: EventFormProps) => {
...
+        <button
+          type="button"
+          onClick={() => setDetailsExpanded((v) => !v)}
+          aria-expanded={detailsExpanded}
+          className="flex w-full items-center justify-between gap-3 rounded-lg border border-blue-200 bg-blue-50 px-4 py-3 text-left shadow-sm transition-colors hover:bg-blue-100"
+        >
+          <div className="min-w-0">
+            <p className="text-[11px] font-medium uppercase tracking-wide text-muted-foreground">
+              {isConnection ? 'Connection' : 'Event'} details
+            </p>
```
alexsapps: thoughts on removing <p>Event details</p> ?
It already says "Event" just above and it's not too had to find out that there are details.

frontend-v2/src/app/(authed)/events/event-form.tsx
```diff
@@ -110,17 +152,55 @@ export const EventForm = ({ mode }: EventFormProps) => {
...
+      eventData.start_time ||
+      eventData.location?.google_place_id ||
+      eventData.description),
+  )
+
+  // When editing a saved event the detail fields (name, type, date, schedule,
+  // location, description) are usually already set, so they collapse behind a
+  // summary header to keep the attendee list near the top for taking
+  // attendance. New events start expanded since you're filling them out.
+  const [detailsExpanded, setDetailsExpanded] = useState(
```
alexsapps: i noticed saving the event collapses the details section. i think this would not be expected, and is annoying if i want to double check the details of what I just saved since not all details are shown when collapsed.

frontend-v2/src/app/(authed)/events/event-form.tsx
alexsapps, line 1: lots going on here. i ask Claude how to refactor :)  it said the below. might be easier to deal w my other comment about putting event details inside the border where the collapse button is if you agree w that.

It's ~950 lines because a single component carries three distinct responsibilities at once:

Data/schema layer — the Zod schema, form types, default-value mapping from the API shape.
Form orchestration — TanStack form setup, the save mutation, the submit/diff logic, derived state subscriptions.
Rendering — every field's JSX inline, plus the collapse/summary UI and attendee section.
The good news: the directory already establishes the right pattern. You've extracted attendee-input-field.tsx, useActivistRegistry.ts, places-autocomplete.tsx. So this is just continuing that decomposition.

Recommended extractions, highest-value first
1. Pull the schema + types into event-form-schema.ts
Lines 39–110 (the EVENT_TYPES const, attendeeSchema, formSchema with its three refinements, FormValues) are pure data with zero React. Moving them out drops ~70 lines and makes the validation rules reviewable on their own. This is the cleanest, lowest-risk cut.

2. Extract a useEventForm hook
The biggest win. Lines ~123–488 are almost entirely non-visual logic that could live in a co-located hook (matching the useActivistRegistry precedent):

the useQuery for existing event
initialValues mapping (266–300)
saveEventMutation with its 90-line onSuccess (176–264)
the form = useForm(...) setup including the submit/diff logic (302–385)
all the useStore subscriptions (396–406)
helpers: checkForDuplicate, ensureMinimumEmptyFields, setDateToToday, the beforeunload effect
The component would then read as: call useEventForm({ mode, startExpanded }), get back { form, ... derived flags ..., handlers }, and render. That alone takes the component from ~950 to ~350–400 lines of mostly JSX.

3. Break the JSX into field sub-components
The render tree has natural seams that map to clear units:

EventDetailsSummaryBar (523–552) — the collapsible header.
ScheduledEventFields (681–850) — start/end time, timezone, online checkbox, location, description. This is the densest block (~170 lines) and is self-contained behind showUpcomingFields.
AttendeesSection (881–921) — the array field + empty-state.
Optionally a small SuppressSurveyField (852–871).
4. Two repeated micro-patterns worth a tiny helper
The error paragraph {field.state.meta.errors[0] && (`<p className="text-sm text-red-500">`...) appears ~6 times. A `<FieldError error={...} />` component removes a lot of noise.
The clear-location block (setFieldValue('googlePlaceId','') … lat/lng to undefined) is duplicated verbatim at 770–776 and 815–821. A single clearLocation(form) helper would collapse both.

frontend-v2/src/app/(authed)/events/event-form.tsx
```diff
@@ -349,161 +516,422 @@ export const EventForm = ({ mode }: EventFormProps) => {
...
-                  ))}
-                </SelectContent>
-              </Select>
-              {field.state.meta.errors[0] && (
-                <p className="text-sm text-red-500">
-                  {field.state.meta.errors[0]?.message}
-                </p>
-              )}
-            </div>
+          {detailsExpanded ? (
```
alexsapps: i think it's visually confusing that the form details appear below the container / bordered box that shows the summary. how about showing the form instead of the summary, within the container?

frontend-v2/src/app/(authed)/events/event-form.tsx
```diff
@@ -110,17 +152,55 @@ export const EventForm = ({ mode }: EventFormProps) => {
     enabled: !!eventId,
   })
 
+  // The scheduled-event fields (time, location, description) are revealed by the
+  // "Public event" checkbox — see showUpcomingFields below. We also surface them
+  // when editing an event that already carries this data, so any pre-existing
+  // event stays fully editable even if its Public box is unchecked.
+  const editingHasUpcomingData = Boolean(
```
alexsapps: i think it's kind of confusing that these fields disappear when the box is unchecked only in certain circumstances (create event page, or edit page when fields aren't set yet).

perhaps these fields should be visible always so as to allow creating a public event without publishing it yet, e.g. if the start time isn't known yet but we have the description available, or we don't want anyone to know about it until just before it starts, or we aren't sure yet if it's happening.

it's also surprising/annoying that checking the box causes more fields to appear IMO. could the checkbox be inside of an accordion when checked or manually expanded?

frontend-v2/src/app/(authed)/home/home-hub.tsx
```diff
@@ -0,0 +1,224 @@
...
+      </header>
+
+      {/* Today's events for quick attendance entry, plus a single entry point
+          for creating any new event (public or not). */}
+      <section className="rounded-xl border bg-card shadow-sm">
+        <div className="flex items-start justify-between gap-3 border-b p-5">
+          <div>
+            <h2 className="text-lg font-semibold">Today&apos;s events</h2>
+            <p className="mt-1 text-sm text-muted-foreground">
+              Pick an event happening today to enter attendance, or create a new
```
alexsapps: remove this help text imo - this page seems pretty straightforward already

frontend-v2/src/app/(authed)/events/event-form.tsx
```diff
@@ -349,161 +516,422 @@ export const EventForm = ({ mode }: EventFormProps) => {
...
+              </div>
             </div>
-          </div>
-        )}
-      </form.Field>
+          )}
+        </form.Field>
+      ) : (
+        <div className="rounded-md border border-dashed p-4 text-sm text-muted-foreground">
+          Save the event first — you can record attendance afterward.
```
alexsapps: how about "You can record attendance after saving the event." to avoid telling the user what to do?

alexsapps: why is the attendance hidden before the event is saved? i wrote some names in attendance before clicking "Publicly listed event" and those names disappeared, but when I hit save I saw they came back and were still saved.

frontend-v2/src/app/(authed)/events/event-form.tsx
```diff
@@ -349,161 +516,422 @@ export const EventForm = ({ mode }: EventFormProps) => {
...
+                        ))}
+                      </SelectContent>
+                    </Select>
+                    <p className="text-xs text-muted-foreground">
+                      Times are in {timezone.replace(/_/g, ' ')}
+                      {getZoneAbbreviation(eventDate, timezone)
+                        ? ` (${getZoneAbbreviation(eventDate, timezone)})`
+                        : ''}
+                      .
+                    </p>
```
alexsapps: how about showing the abbreviations inside the `SelectItem`s so there's no need to clarify them in a paragraph below?

frontend-v2/src/app/(authed)/events/event-form.tsx
```diff
@@ -66,27 +66,69 @@ const attendeeSchema = z.object({
...
+    path: ['endTime'],
+  })
+  // Publicly listed events are scheduled in advance, so a start time is required.
+  .refine((v) => !v.isPublic || Boolean(v.startTime), {
+    message: 'Start time is required for public events',
+    path: ['startTime'],
+  })
+  // In-person public events need a location; online ones don't.
+  .refine((v) => !v.isPublic || v.isOnline || Boolean(v.googlePlaceId), {
+    message: 'Location is required for in-person public events',
```
alexsapps: often times i imagine locations might not be in Google Places, like if we meet at an intersection or some public land that isn't quite a park, and we just want to provide coordinates instead. i haven't tried with a google places API key yet but does this allow describing the location and choosing coordinates manually? and if we do support that, i imagine we'd want a way to rename locations from past events that have confusing, or overly broad names to make room for new location names. by the way, i'm pretty sure Facebook events and Google calendar let you type whatever you want for their location fields probably because places API is too restrictive.

frontend-v2/src/app/(authed)/events/[id]/confirmation/event-confirmation.tsx
```diff
@@ -0,0 +1,103 @@
...
+    event?.location?.name || event?.location?.formatted_address
+
+  return (
+    <div className="flex flex-col gap-6">
+      <div className="flex flex-col items-center gap-3 text-center">
+        <CheckCircle2 className="h-12 w-12 text-green-600" />
+        <div className="flex flex-col gap-1">
+          <h2 className="text-xl font-semibold">Event created</h2>
+          <p className="text-sm text-muted-foreground">
+            Your event is scheduled. Take attendance when the event happens.
```
Comment (alexsapps) [inline frontend-v2/src/app/(authed)/events/[id]/confirmation/event-confirmation.tsx]: "Take attendance when the event happens." -- yeah we know x) pls remove

frontend-v2/src/app/(authed)/events/event-form.tsx
```diff
@@ -186,8 +282,22 @@ export const EventForm = ({ mode }: EventFormProps) => {
           : Array(DEFAULT_FIELD_COUNT)
               .fill(null)
               .map(() => ({ name: '' })),
+      // Scheduled-event fields. Public defaults off, so a new event starts as the
+      // quick attendance form; checking "Public event" reveals the rest.
+      isPublic: eventData?.is_public ?? false,
+      isOnline: eventData?.is_online ?? false,
+      description: eventData?.description ?? '',
+      // MySQL returns TIME as "HH:MM:SS"; slice to the "HH:MM" the input expects.
```
alexsapps: frontend should not know "MySQL", and should backend trim the seconds instead? also you have a `toTime` function for this i think you could use.

frontend-v2/src/app/(authed)/events/event-form.tsx
```diff
@@ -237,6 +358,28 @@ export const EventForm = ({ mode }: EventFormProps) => {
         added_attendees: addedAttendees,
         deleted_attendees: deletedAttendees,
         suppress_survey: value.suppressSurvey,
+        // Only send scheduled-event fields when they're in play, keeping the
```
alexsapps: premature optimization?

frontend-v2/src/lib/time.ts
```diff
@@ -0,0 +1,186 @@
...
+  if (!start) return ''
+  const end = formatWallClock(event.end_time ?? '')
+  const abbr = getZoneAbbreviation(event.event_date, event.timezone ?? '')
+  const range = end ? `${start} – ${end}` : start
+  return abbr ? `${range} ${abbr}` : range
+}
+
+/** Parses "HH:MM" or "HH:MM:SS" into minutes since midnight; null if invalid. */
+function toMinutesSinceMidnight(time: string): number | null {
+  const match = /^(\d{1,2}):(\d{2})/.exec(time.trim())
```
alexsapps: can you reuse `toTime` here?

frontend-v2/src/lib/time.ts
```diff
@@ -0,0 +1,186 @@
...
+  }
+}
+
+/**
+ * Whether `now` falls within an event's [start, end) window, evaluated in the
+ * event's own timezone. Events without an end time use a 1-hour default
+ * duration. Requires a start time and the event being on its own local "today";
+ * returns false otherwise.
+ */
+export function isEventHappeningNow(
```
alexsapps, line 183: this function isn't timezone specific. should this file just be called time.ts, or perhaps this function should be moved to the call site?

server/src/model/event.go
```diff
@@ -512,9 +604,138 @@ func CleanEventData(db *sqlx.DB, body io.Reader, chapterID int) (Event, error) {
...
+
+	// A publicly listed event must form a coherent schedule. The form enforces
+	// this, but guard server-side too so a malformed payload can't create a
+	// broken public listing. (end_time before start_time is intentionally not
+	// rejected here — see the form's note on overnight events.)
+	if e.IsPublic {
+		if !e.StartTime.Valid {
+			return Event{}, fmt.Errorf("public events require a start time")
```
alexsapps: i've been using `ValidationErrorf` to indicate errors that are not re-tryable. would be nice if you could use it for such errors going forward so we can return http 400 instead of 500. i ought to mention it in claude.md sometime.

server/src/model/event.go
```diff
@@ -512,9 +604,138 @@ func CleanEventData(db *sqlx.DB, body io.Reader, chapterID int) (Event, error) {
...
+		return 0, fmt.Errorf("failed to create transaction: %w", err)
+	}
+
+	_, err = tx.Exec(`INSERT INTO locations (chapter_id, google_place_id, name, formatted_address, lat, lng)
+VALUES (?, ?, ?, ?, ?, ?)
+ON DUPLICATE KEY UPDATE
+  name = IF(VALUES(name) <> '', VALUES(name), name),
+  formatted_address = IF(VALUES(formatted_address) <> '', VALUES(formatted_address), formatted_address),
+  lat = COALESCE(VALUES(lat), lat),
+  lng = COALESCE(VALUES(lng), lng)`,
```
alexsapps: should these values be updated too if they change? (i.e. moved into "ON DUPLICATE KEY UPDATE")

it's probably not great to change the location of past events, but probably better not to use an outdated location for an upcoming event.

alexsapps (changes requested):
exciting times. thanks Jake!

frontend-v2/src/app/(authed)/home/home-hub.tsx
jakehobbs: Keeping this at `/home` — it's the post-login landing page (the logo links here), and everything it offers is reachable from the events pages too, so I'd rather not surface 'home'/'today' as a nav concept. I did make it earn its keep though: the events content now only renders for users with attendance access, so others just get the header. Leaving the greeting + date as a light orienting touch.

(posted by claude)

jakehobbs: The Attendance menu already has 'New Event' and 'All Events', which cover what this page links to, so a dedicated entry would mostly duplicate them — and I'd rather not add a 'Home'/'Today' label to the nav. On the role question: the page stays everyone's landing page, but its events section now only shows for users with attendance access (admin/organizer/attendance), so people without it just see the header.

(posted by claude)

server/src/main.go
jakehobbs: Removed — the note now lives where the key is defined in config.go.

(posted by claude)

frontend-v2/src/app/authed-page-provider.tsx
jakehobbs: Removed.

(posted by claude)

Makefile
jakehobbs: Left this one: the Go Server debug config loads a single `envFile` (`server/debug.env`), and the new `.env` is optional (only needed for the Places key locally). Pointing the debugger at a `.env` most people won't have would break 'Go Server' for them. If you want Places in the debugger, easiest is to drop `GOOGLE_PLACES_API_KEY` into your local `server/debug.env`. I documented the `.env` route in the README.

(posted by claude)

jakehobbs: Done — added a note under the Google Places section of the README on populating a root `.env`, including that the key is in GCP under the dev environment.

(posted by claude)

frontend-v2/src/components/nav.tsx
jakehobbs: I'd rather not gate it: organizers should start creating these events now, ahead of the public listing going live, so the data's there when it does. Nothing here is publicly exposed yet — it's admin-only scheduling + attendance — so there's no risk shipping it un-flagged.

(posted by claude)

frontend-v2/src/app/(authed)/events/event-form.tsx
jakehobbs: Done, removed it.

(posted by claude)

frontend-v2/src/app/(authed)/events/event-form.tsx
jakehobbs: Good catch — stopped force-collapsing the details on save, so it keeps whatever state you had.

(posted by claude)

frontend-v2/src/app/(authed)/events/event-form.tsx
jakehobbs: Done — split it up: `event-form-schema.ts` (schema/types), a `useEventForm` hook for the non-visual logic, and `event-form-sections.tsx` for the JSX sub-sections (summary bar, scheduled fields, attendees), plus a shared `FieldError` component and a `clearLocation` helper for the duplicated blocks. The component is down to ~310 lines of mostly JSX.

(posted by claude)

frontend-v2/src/app/(authed)/events/event-form.tsx
jakehobbs: I didn't move the whole form inside the summary box, but I tightened the grouping: the scheduled-event fields are now wrapped in a nested bordered sub-box so they read as one revealed section rather than loose fields below. Kept the summary bar as the collapse control for existing events.

(posted by claude)

frontend-v2/src/app/(authed)/events/event-form.tsx
jakehobbs: I want to keep the 'Publicly listed event' checkbox as the gate — always showing these would clutter the quick-attendance flow most people use. Didn't go to a full accordion, but the fields are now grouped in a nested sub-box so the reveal reads as a clear, contained section rather than fields popping inline.

(posted by claude)

frontend-v2/src/app/(authed)/home/home-hub.tsx
jakehobbs: Removed.

(posted by claude)

frontend-v2/src/app/(authed)/events/event-form.tsx
jakehobbs: Done — now reads 'You can record attendance after saving the event.'

(posted by claude)

jakehobbs: Intentional for the create-in-advance flow: a brand-new public event is saved first and attendance is recorded later (at the event), so hiding the attendee inputs keeps the create form simpler. The names aren't lost when you toggle the box — they're kept in form state and come back on save, which is the behavior you saw.

(posted by claude)

frontend-v2/src/app/(authed)/events/event-form.tsx
jakehobbs: Done — each timezone option now shows its abbreviation, e.g. 'America/Los Angeles (PDT)', and I dropped the redundant 'Times are in…' line below.

(posted by claude)

frontend-v2/src/app/(authed)/events/event-form.tsx
jakehobbs: Agreed, this is a real gap — lots of spots aren't clean Google Places. Going ahead and building it: an 'advanced' manual-location mode (free-text name + optional coordinates, stored on the event rather than the place-keyed `locations` table). Coming in a follow-up commit on this PR. Renaming existing location names I'll leave for later since locations aren't surfaced in the UI yet.

(posted by claude)

frontend-v2/src/app/(authed)/events/[id]/confirmation/event-confirmation.tsx
Comment (jakehobbs) [reply frontend-v2/src/app/(authed)/events/[id]/confirmation/event-confirmation.tsx]: Removed.

(posted by claude)

frontend-v2/src/app/(authed)/events/event-form.tsx
jakehobbs: Good call — moved the trim to the backend: `EventJSON` now returns `start_time`/`end_time` as 'HH:MM' (the API contract), so the frontend no longer slices or knows anything about MySQL's 'HH:MM:SS'. (`toTime` is a private helper inside the react-aria TimeField that returns a `Time` object, so it wasn't reusable here — but the slice is gone either way.)

(posted by claude)

frontend-v2/src/app/(authed)/events/event-form.tsx
jakehobbs: It's a guard rather than an optimization: the conditional keeps the connection and quick-attendance save payloads exactly as they were (no `is_public`/time/location fields), so those flows — coaching especially, which hits a different endpoint — don't start sending event-only fields. Always including them risks clobbering/confusing those paths, so I'd keep it.

(posted by claude)

frontend-v2/src/lib/time.ts
jakehobbs: The only `toTime` is a private helper inside the react-aria `TimeField` and returns an `@internationalized/date` `Time` object; `toMinutesSinceMidnight` needs an integer for the 'happening now' math, so reusing it would mean pulling a UI-layer dependency into a pure util. Kept them separate.

(posted by claude)

frontend-v2/src/lib/time.ts
jakehobbs: Done — renamed it to `time.ts` and updated the imports.

(posted by claude)

server/src/model/event.go
jakehobbs: Done — switched these to `ValidationErrorf` (and the start/end-time + timezone parse errors too) so bad input returns a 400 instead of a 500. Thanks for the pointer.

(posted by claude)

server/src/model/event.go
jakehobbs: These are already updated on conflict — the `ON DUPLICATE KEY UPDATE` refreshes name/address/lat/lng, guarded so a blank/missing value in a sparser response can't wipe a richer stored one. So an upcoming event that re-picks an existing place gets the latest non-empty metadata.

(posted by claude)

frontend-v2/src/app/(authed)/events/event-form.tsx
alexsapps: how about guarding with just "!isConnection" and removing the comment? it's just a lot to understand right now with the shape of the json depending on certain fields being set.

Makefile
alexsapps: so far makefile and launch.json behavior have been kept in sync and i'd like to keep it that way if possible, so we don't have situations where things work depending on how they're launched. perhaps we could have 'make deps' create the file if it's missing so it can be added to launch.json too?

frontend-v2/src/components/nav.tsx
alexsapps: should we at least add some clarifying help text that says that events are not actually published anywhere yet?

frontend-v2/src/app/(authed)/events/event-form.tsx
alexsapps: thinking about this more, i think there are two things strange about this specifically:
1. attendance appears and disappears depending on whether the event is saved / public. i think the intent is to avoid confusing people with attendance fields for events that are happening in the future. in that case, why not look at the event date to determine if attendance should be available?
2. saving fields that are not present. what if someone does take attendance with some names for a new future event, then marks it as public and feels no need to delete those names because they've disappeared? it may be confusing for the future attendance taker to know what to do when they see those names show up later. this could be fixed by clearing / not setting attendance for new events that are in the future matching what the UI implies.

frontend-v2/src/components/nav.tsx
```diff
@@ -282,9 +282,13 @@ export const Navbar = () => {
...
+          href="/home"
```
alexsapps, line 286: please add in Vue nav component too

frontend-v2/src/app/(authed)/home/home-hub.tsx
alexsapps: just for the boilerplate then, how about putting the chapter name and today's date underneath "Today's events" and removing "Welcome, name"? i feel like the current boilerplate is a little cramped and little cheesy lol.

frontend-v2/src/app/(authed)/events/event-form.tsx
alexsapps: But won't block on it but I still think the form should be in the same container where the arrow is, or perhaps the details and attendance should be separated in some other way. Maybe can revisit later if adding more fields later like photo upload. (The fix you mentioned is good but not related to this, I think.)

alexsapps: Btw Katya thought putting the attendance in an accordion too could make sense, in case there are hundreds of people in attendance. That would making clicking save easier after changing event details. But we can get feedback from real users first before worrying about it.

Makefile
alexsapps: Is that actually where the key you used is? I tried it and got this error:

Google Maps JavaScript API error: ApiNotActivatedMapError
https://developers.google.com/maps/documentation/javascript/error-messages#api-not-activated-map-error

Would be great to mention which project/key in the readme. I'd like to test it out so pls lmk :)

frontend-v2/src/app/(authed)/events/event-form.tsx
alexsapps: how about gating this PR's changes, or even just visibility of the location field, behind a feature flag, just until we can rename places? i could see it becoming a pain point very quickly that a place can't be renamed. i don't want to get tech requests to fix typos in place names lol.

frontend-v2/src/app/(authed)/home/home-hub.tsx
jakehobbs: yep sounds good, doing that

Makefile
jakehobbs: not the biggest fan of dev containers. haha. lemme see what i can do.

jakehobbs: i think i figured it out

jakehobbs: done. sounds like would be good for us to use secrets manager.

frontend-v2/src/components/nav.tsx
jakehobbs: 🤔 hmmmm was planning to send people an email or something

jakehobbs: let me see how a tooltip would look

frontend-v2/src/app/(authed)/events/event-form.tsx
jakehobbs: struggling to come up w/ a solution i'm totally satisfied with. i feel like we need a design for stuff like this before implement, ideally. haha. i think just shipping like this for MVP and reiterating later makes sense tho.

frontend-v2/src/app/(authed)/events/event-form.tsx
jakehobbs: made some tweaks

frontend-v2/src/app/(authed)/events/event-form.tsx
jakehobbs: what do you mean by renaming places? i let you type in whatever name you want & give coords now.

frontend-v2/src/app/(authed)/events/event-form.tsx
jakehobbs: Done — switched the spread to guard on just \`!isConnection\` and removed the comment. It is behavior-equivalent: the backend nulls out the empty scheduled fields for plain attendance saves (\`cleanEventTime("")\` → NULL, and the \`is_public\`-requires-start-time guard only fires when \`is_public\` is true), and the form always loads these fields into state, so nothing gets clobbered.

Kept the \`includeUpcoming\` variable itself though — it is still load-bearing for the "at least one attendee required" check above, since public/upcoming events legitimately have no attendees yet.

(Heads up: this block moved to \`useEventForm.ts\` since you commented.)

— posted by claude 🤖

frontend-v2/src/components/nav.tsx
jakehobbs: done

frontend-v2/src/app/(authed)/events/event-form.tsx
jakehobbs: oh, do you mean google places? hmmm, i'm reworking this a bit.

jakehobbs: Ah, I think we were talking about two different things. The manual free-text path I added does let you type any name + coords. The gap you're pointing at is the **Google Places** path: when you picked a place, its name came from Google and got stored in a shared `locations` table (keyed by place id) — so that name wasn't editable, and it was shared across every event using that place. That's the "can't rename" pain you were worried about, and renaming it centrally would've silently changed other events too.

Rather than feature-flag it, I restructured so this can't happen: there's no shared locations table anymore. Every event stores its own free-text location **name** (always editable), plus optional geo data — a Google place id and/or coordinates — used just for the map pin/address. Picking a Google place now only prefills the name (if you haven't typed one) and fills the coords; you can edit the name freely afterward.

So: type whatever you want for the name (like FB/Google Calendar), fix typos per-event without affecting anything else, and no feature flag needed. Pushed in the latest commits.

(posted by claude)

Makefile
alexsapps: ✅ still wasn't working for me so i'm pushing some commits to your branch to fix 🙂

btw in theory launch.json has nothing to do with devcontainers, although maybe there is some dependency with our specific launch.json that you encountered.

alexsapps: ✅ ok ty! just a couple notes:

for prod, perhaps in the future (if this is a prod key) we should keep all the ADB keys in the ADB project, otherwise it's hard to know when a key can be deleted, and probably hard to see how much each app is consuming/costing us in GCP if they share projects or are split between projects. maybe i'll move it later if you don't object. although there are two ADB projects in GCP...which complicates things lol.

for dev, in the future, we can use the "DxE General Dev Test" project. this makes clear that it's safer to give the key to an untrusted contributor.

frontend-v2/src/app/(authed)/events/event-form.tsx
alexsapps: ✅ ty!

frontend-v2/src/app/(authed)/events/event-form.tsx
alexsapps: ✅ sg 🙂

frontend-v2/src/app/(authed)/events/event-form.tsx
alexsapps: ✅ ty!

alexsapps (dismissed):
non-blocking issues claude found which i may look into later if i have time:

~~Duplicated parent→local value sync pattern — places-autocomplete.tsx
PlacesAutocomplete syncs parent-driven changes with local state to avoid flashing stale text; home-hub.tsx does a similar sync via useSyncExternalStore (for hydration safety). The finder suggested extracting the shared pattern into a reusable hook to avoid future sync bugs.~~

FieldError component not used everywhere — event-form-sections.tsx
A new FieldError component was introduced for error messages, but attendee-input-field.tsx still renders errors inline with the same text-sm text-red-500 mt-1 classes. So error styling now lives in two places; changing it means editing both.

toMinutesSinceMidnight duplicates formatWallClock's parsing — time.ts
Both functions independently parse the same HH:MM regex. If the accepted time format ever changes, both must be updated. Suggested extracting a single parse helper.

~~setDateToToday efficiency nit — useEventForm.ts
Minor: reuse the already-memoized browserTz rather than recomputing browser-timezone detection. Low impact.~~

alexsapps:
> non-blocking issues claude found which i may look into later if i have time:
> 
> ~Duplicated parent→local value sync pattern — places-autocomplete.tsx PlacesAutocomplete syncs parent-driven changes with local state to avoid flashing stale text; home-hub.tsx does a similar sync via useSyncExternalStore (for hydration safety). The finder suggested extracting the shared pattern into a reusable hook to avoid future sync bugs.~
> 
> FieldError component not used everywhere — event-form-sections.tsx A new FieldError component was introduced for error messages, but attendee-input-field.tsx still renders errors inline with the same text-sm text-red-500 mt-1 classes. So error styling now lives in two places; changing it means editing both.
> 
> toMinutesSinceMidnight duplicates formatWallClock's parsing — time.ts Both functions independently parse the same HH:MM regex. If the accepted time format ever changes, both must be updated. Suggested extracting a single parse helper.
> 
> ~setDateToToday efficiency nit — useEventForm.ts Minor: reuse the already-memoized browserTz rather than recomputing browser-timezone detection. Low impact.~

pushed commits to fix 2 of them. other 2 not worth fixing. feel free to revert if i'm missing anything with any of my commits.

PR #430, opened 2026-05-31
fix(frontend-v2): pin eslint to v9 for plugin compatibility

jakehobbs:
ESLint was bumped to 10.4.0 by dependabot (commit 0ff3e834), but ESLint 10 removed `context.getFilename()`, which `eslint-plugin-react@7.37.5` (pulled in transitively by `eslint-config-next@16.2.6`) still calls — crashing every lint run with `contextOrFilename.getFilename is not a function`. This is a [known upstream issue](https://github.com/vercel/next.js/issues/89764) with no fixed `eslint-plugin-react`/`eslint-config-next` release yet, and the maintainer-tracked recommendation is to stay on ESLint v9 (which `eslint-config-next@16.2.6` fully supports). This PR pins `eslint` back to `^9.39.4` so linting works again; we can re-bump to v10 once upstream ships compatibility.

## Summary by CodeRabbit

* **Chores**
  * Updated development dependencies and tooling versions
  * Enhanced code quality enforcement with stricter linting validation

* **Refactor**
  * Optimized internal state management in multiple components to improve rendering efficiency

### DIFF
--- a/frontend-v2/package.json
+++ b/frontend-v2/package.json
@@ -54,7 +54,7 @@
     "@types/node": "^25.9.1",
     "@types/react": "^19.2.15",
     "@types/react-dom": "^19.2.3",
-    "eslint": "^10.4.0",
+    "eslint": "^9.39.4",
     "eslint-config-next": "16.2.6",
     "eslint-plugin-react-hooks": "^7.1.1",
     "jsdom": "^29.1.1",
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -12,6 +12,13 @@ updates:
     directory: /frontend-v2
     schedule:
       interval: weekly
+    ignore:
+      # Pin ESLint to the v9 line: Next.js isn't compatible with ESLint 10 yet.
+      - dependency-name: "eslint"
+        update-types: ["version-update:semver-major"]
+      # Don't bump Tailwind to v4+: it requires a manual migration (v3 -> v4).
+      - dependency-name: "tailwindcss"
+        update-types: ["version-update:semver-major"]
     groups:
       all-npm:
         patterns:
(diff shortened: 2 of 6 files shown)


jakehobbs:
## Review summary

I reviewed the workspace diff: an eslint downgrade (`^10.4.0` → `^9.39.4`, plus lockfile) and two source files updated to satisfy `react-hooks` rules that the version change surfaced.

**`date-range-filter.tsx`** — Replaces a `useEffect`-based prop→state sync with React's documented "adjust state during render" pattern (`prevUnit` guard + `setLocalUnit`). Behavior is equivalent to the old effect (resets `localUnit` only when the `unit` prop actually changes; user-driven changes via `handleUnitChange` are preserved), and it avoids the extra render/flash. Correct.

**`useActivistRegistry.ts`** — Replaces `useRef(new ActivistRegistry())` with a `useState(() => new ActivistRegistry())` lazy initializer. The instance is still stable for the component's lifetime, it's now legitimately readable during render, and the lazy initializer avoids allocating a throwaway `ActivistRegistry` on every render that the old `useRef` argument incurred. The added `registry` entries in the effect dependency arrays are no-ops since the value is stable. The two `react-hooks/set-state-in-effect` disables are correctly scoped to genuine one-shot loading flags. Behavior-preserving.

**No findings to flag.** The changes are correct, behavior-preserving, and well-commented. The eslint pin to v9 is reasonable given `eslint-plugin-react-hooks`/`eslint-config-next` compatibility. I did not post any inline comments.

alexsapps (dismissed):

frontend-v2/package.json
```diff
@@ -54,7 +54,7 @@
     "@types/node": "^25.9.1",
     "@types/react": "^19.2.15",
     "@types/react-dom": "^19.2.3",
-    "eslint": "^10.4.0",
```
alexsapps, line 57: i'll just try to remember not to let this happen again for now. but someday we should run the linter as part of the prod build.

alexsapps (commented):
ty!!!

frontend-v2/package.json
alexsapps: nevermind i updated dependabot config for this. pushed to your branch here.

PR #455, opened 2026-07-06
fix(devcontainer): make worktree script not break windows users

alexsapps:
but don't actually support worktrees on Windows yet since Windows users tend to be less involved and probably won't need it.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved workspace setup for Windows paths, making container-based development more reliable on Windows hosts.
  * Updated path handling to better support workspace mounting and naming across different path formats.
  * Adjusted worktree handling so Windows setups skip unsupported extra mount logic, avoiding setup failures.


alexsapps:
@coderabbitai review

PR #462, opened 2026-07-19
Port external events admin page to React

jakehobbs:
## Summary
- Ports the admin-only "Facebook Events" moderation page from Vue (`frontend/FacebookEvents.vue`, route `/admin/external_events`) to Next.js at `frontend-v2/src/app/(authed)/external-events/page.tsx`, served at `/v2/external-events`.
- Lists upcoming Facebook/Eventbrite events for the SF Bay page (same hardcoded page ID and "today onward" window as the legacy page), with actions to feature/unfeature an event or cancel (soft-delete) it.
- No backend changes; the legacy Vue page and its route are untouched and remain available under "Legacy" in the nav.

## Endpoints used (all pre-existing, no Go changes)
- `GET /external_events/{page_id}?start_time=...` — lists events (same endpoint the Vue page calls; not itself admin-gated at the API level, but the new route/nav entry are admin-only).
- `POST /admin/external_events/feature` — `{ id, featured }`, toggles the featured flag.
- `POST /admin/external_events/cancel` — `{ id }`, soft-deletes (sets `is_canceled`) an event.

Added typed `ApiClient` methods (`getExternalEvents`, `featureExternalEvent`, `cancelExternalEvent`) and `API_PATH` entries in `frontend-v2/src/lib/api.ts`, following the file's existing zod-schema + `handleKyError`/`throwIfApiError` conventions. Note the Go `ExternalEvent` struct has no `json` tags, so response fields are PascalCase (`ID`, `Name`, `StartTime`, `Featured`) — `ID` is a string (Facebook IDs), not a number.

- Admin dropdown's "Facebook Events" item now points to `/v2/external-events` with `page: "FacebookEvents_beta"` (still `roleRequired: ["admin"]`).
- Appended a "Facebook Events" entry to the end of the "Legacy" dropdown pointing at the old `/admin/external_events` route, so the Vue page stays reachable.

## Feature parity / intentional differences
- Full parity: list events, format start date, link event name to `facebook.com/{id}`, toggle featured (with success/error toast instead of the Vue `flashMessage`), cancel event with confirmation.
- Added a confirmation dialog before cancelling an event (the Vue page cancelled immediately on click with no confirmation) — matches this app's existing destructive-action pattern (see `hide-activist-dialog.tsx`).
- Admin gating mirrors `frontend-v2/src/app/(authed)/config/page.tsx`: server component checks `session.user.Roles.includes('admin')` and calls `forbidden()` otherwise.

## Test plan
- [x] `pnpm exec tsc --noEmit` — no new errors (repo has 2 pre-existing/environmental errors unrelated to this change: `tsconfig.json` `baseUrl` deprecation warning under the installed TypeScript version, and a missing `$public/logo.png` module type in `nav.tsx`; both reproduce identically on `main` before this change).
- [x] `pnpm lint` — clean.
- [x] `pnpm build` — succeeds, `/external-events` route builds.
- [ ] Manually verify in a running dev environment: load `/v2/external-events` as an admin, feature/unfeature and cancel an event, and confirm a non-admin is forbidden.

### DIFF
--- a/frontend-v2/src/app/(authed)/external-events/cancel-event-dialog.tsx
+++ b/frontend-v2/src/app/(authed)/external-events/cancel-event-dialog.tsx
@@ -54,7 +54,7 @@ export function CancelEventDialog({
         </DialogHeader>
         <p className="text-sm">
           Are you sure you want to cancel &ldquo;{eventName}&rdquo;? It will no
-          longer be shown to activists.
+          longer be displayed on the public events page.
         </p>
         <DialogFooter>
           <Button
--- a/frontend-v2/src/app/(authed)/external-events/cancel-event-dialog.tsx
+++ b/frontend-v2/src/app/(authed)/external-events/cancel-event-dialog.tsx
@@ -1,0 +1,80 @@
+'use client'
+
+import { useMutation, useQueryClient } from '@tanstack/react-query'
+import toast from 'react-hot-toast'
+import { API_PATH, apiClient } from '@/lib/api'
+import { Button } from '@/components/ui/button'
+import {
+  Dialog,
+  DialogContent,
+  DialogFooter,
+  DialogHeader,
+  DialogTitle,
+} from '@/components/ui/dialog'
+
+type Props = {
+  open: boolean
+  onOpenChange: (open: boolean) => void
+  eventId: string
+  eventName: string
+}
+
+export function CancelEventDialog({
+  open,
+  onOpenChange,
+  eventId,
+  eventName,
+}: Props) {
+  const queryClient = useQueryClient()
+
+  const mutation = useMutation({
+    mutationFn: () => apiClient.cancelExternalEvent(eventId),
+    onSuccess: () => {
+      toast.success('Successfully cancelled event.')
+      queryClient.invalidateQueries({
+        queryKey: [API_PATH.EXTERNAL_EVENTS_LIST],
+      })
+      onOpenChange(false)
+    },
+    onError: (err: Error) => {
+      toast.error(err.message || 'Failed to cancel event')
+    },
+  })
+
+  const handleOpenChange = (next: boolean) => {
+    if (mutation.isPending) return
+    onOpenChange(next)
+  }
+
+  return (
+    <Dialog open={open} onOpenChange={handleOpenChange}>
+      <DialogContent>
+        <DialogHeader>
+          <DialogTitle>Cancel event</DialogTitle>
+        </DialogHeader>
+        <p className="text-sm">
+          Are you sure you want to cancel &ldquo;{eventName}&rdquo;? It will no
+          longer be shown to activists.
+        </p>
+        <DialogFooter>
+          <Button
+            type="button"
+            variant="outline"
+            onClick={() => handleOpenChange(false)}
+            disabled={mutation.isPending}
+          >
+            Keep event
+          </Button>
+          <Button
+            type="button"
+            variant="destructive"
+            onClick={() => mutation.mutate()}
+            disabled={mutation.isPending}
+          >
+            {mutation.isPending ? 'Cancelling...' : 'Cancel event'}
+          </Button>
+        </DialogFooter>
+      </DialogContent>
+    </Dialog>
+  )
+}
--- a/frontend-v2/src/app/(authed)/external-events/external-events-page.tsx
+++ b/frontend-v2/src/app/(authed)/external-events/external-events-page.tsx
@@ -1,0 +1,40 @@
+'use client'
+
+import { useQuery } from '@tanstack/react-query'
+import { Loader2 } from 'lucide-react'
+import { API_PATH, apiClient } from '@/lib/api'
+import { ExternalEventsTable } from './external-events-table'
+
+export default function ExternalEventsPage() {
+  const {
+    data: events,
+    isLoading,
+    isError,
+  } = useQuery({
+    queryKey: [API_PATH.EXTERNAL_EVENTS_LIST],
+    queryFn: ({ signal }) => apiClient.getExternalEvents(signal),
+  })
+
+  return (
+    <div className="flex flex-col gap-4">
+      <div className="flex flex-col gap-1">
+        <h1 className="text-2xl font-semibold">Facebook Events</h1>
+        <p className="text-muted-foreground text-sm">
+          Feature or cancel upcoming Facebook/Eventbrite events shown to
+          activists.
+        </p>
+      </div>
+
+      {isLoading ? (
+        <div className="flex items-center gap-2 text-sm text-muted-foreground">
+          <Loader2 className="h-4 w-4 animate-spin" />
+          Loading events...
+        </div>
+      ) : isError || !events ? (
+        <div className="text-sm text-destructive">Failed to load events</div>
+      ) : (
+        <ExternalEventsTable events={events} />
+      )}
+    </div>
+  )
+}
--- a/frontend-v2/src/app/(authed)/external-events/external-events-page.tsx
+++ b/frontend-v2/src/app/(authed)/external-events/external-events-page.tsx
@@ -6,11 +6,7 @@ import { API_PATH, apiClient } from '@/lib/api'
 import { ExternalEventsTable } from './external-events-table'
 
 export default function ExternalEventsPage() {
-  const {
-    data: events,
-    isLoading,
-    isError,
-  } = useQuery({
+  const { data: events, isLoading } = useQuery({
     queryKey: [API_PATH.EXTERNAL_EVENTS_LIST],
     queryFn: ({ signal }) => apiClient.getExternalEvents(signal),
   })
@@ -20,8 +16,8 @@ export default function ExternalEventsPage() {
       <div className="flex flex-col gap-1">
         <h1 className="text-2xl font-semibold">Facebook Events</h1>
         <p className="text-muted-foreground text-sm">
-          Feature or cancel upcoming Facebook/Eventbrite events shown to
-          activists.
+          Feature or cancel upcoming Facebook/Eventbrite events displayed on the
+          public events page.
         </p>
       </div>
 
@@ -30,10 +26,10 @@ export default function ExternalEventsPage() {
           <Loader2 className="h-4 w-4 animate-spin" />
           Loading events...
         </div>
-      ) : isError || !events ? (
-        <div className="text-sm text-destructive">Failed to load events</div>
-      ) : (
+      ) : events ? (
         <ExternalEventsTable events={events} />
+      ) : (
+        <div className="text-sm text-destructive">Failed to load events</div>
       )}
     </div>
   )
--- a/frontend-v2/src/app/(authed)/external-events/external-events-table.tsx
+++ b/frontend-v2/src/app/(authed)/external-events/external-events-table.tsx
@@ -1,0 +1,225 @@
+'use client'
+
+import { useMemo, useState } from 'react'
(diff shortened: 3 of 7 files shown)


frontend-v2/src/app/(authed)/external-events/page.tsx
```diff
@@ -0,0 +1,38 @@
...
+  const queryClient = new QueryClient()
+
+  // Intentionally use prefetchQuery here: this endpoint isn't itself
+  // admin-gated (the legacy Vue page calls it unauthenticated), so there's no
+  // 403/404 to redirect on and a failed prefetch can just fall through to the
+  // client-side useQuery below.
+  await queryClient.prefetchQuery({
+    queryKey: [API_PATH.EXTERNAL_EVENTS_LIST],
+    queryFn: ({ signal }) => apiClient.getExternalEvents(signal),
+  })
```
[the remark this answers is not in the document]
jakehobbs: [Claude] Not changed — I disagree with this one. TanStack Query v5's `prefetchQuery` never rejects; it internally swallows errors (`.catch(noop)`) precisely so a failed prefetch doesn't fail the server render. The repo's own `server-auth.ts` documents this exact distinction vs `fetchQuery` (which does rethrow). The comment in `page.tsx` is accurate as written.

frontend-v2/src/lib/api.ts
```diff
@@ -750,6 +777,56 @@ export class ApiClient {
       return this.handleKyError(err)
     }
   }
+
+  // Lists upcoming Facebook/Eventbrite events for the SF Bay page, matching
+  // the legacy admin FacebookEvents.vue page's hardcoded page ID and
+  // "today onward" start_time window.
+  getExternalEvents = async (signal?: AbortSignal) => {
+    try {
+      const startTime = `${format(new Date(), 'yyyy-MM-dd')}T00:00:00Z`
```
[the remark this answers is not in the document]
jakehobbs: [Claude] Fixed in 2b286d64: `start_time` is now derived from `toISOString()` (UTC) instead of the host's local time via `format(new Date(), ...)`, so the computed day boundary is consistent between SSR and the client.

frontend-v2/src/app/(authed)/external-events/external-events-table.tsx
```diff
@@ -0,0 +1,225 @@
...
+            <Trash2 />
+            Delete
+          </Button>
+        ),
+      },
+    ],
+    [featureMutation],
+  )
+
+  // eslint-disable-next-line react-hooks/incompatible-library -- Remove once TanStack Table supports React Compiler.
```
jakehobbs, line 151: why do we need this comment? is it used on other pages?

frontend-v2/src/app/(authed)/external-events/cancel-event-dialog.tsx
```diff
@@ -0,0 +1,80 @@
...
+import {
+  Dialog,
+  DialogContent,
+  DialogFooter,
+  DialogHeader,
+  DialogTitle,
+} from '@/components/ui/dialog'
+
+type Props = {
+  open: boolean
```
jakehobbs, line 16: is `open` the pattern other dialogs use? i would prefer `isOpen`, but don't change it if this is the established pattern.

frontend-v2/src/app/(authed)/external-events/cancel-event-dialog.tsx
```diff
@@ -0,0 +1,80 @@
...
+        queryKey: [API_PATH.EXTERNAL_EVENTS_LIST],
+      })
+      onOpenChange(false)
+    },
+    onError: (err: Error) => {
+      toast.error(err.message || 'Failed to cancel event')
+    },
+  })
+
+  const handleOpenChange = (next: boolean) => {
```
jakehobbs, line 44: is this an existing pattern exactly in this codebase? if not, pls rename `next` to `nextState`.

frontend-v2/src/app/(authed)/external-events/cancel-event-dialog.tsx
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  return (
+    <Dialog open={open} onOpenChange={handleOpenChange}>
+      <DialogContent>
+        <DialogHeader>
+          <DialogTitle>Cancel event</DialogTitle>
+        </DialogHeader>
+        <p className="text-sm">
+          Are you sure you want to cancel &ldquo;{eventName}&rdquo;? It will no
```
jakehobbs, line 56: change to say it will no longer be displayed on the public events page

frontend-v2/src/app/(authed)/external-events/external-events-page.tsx
```diff
@@ -0,0 +1,40 @@
...
+    queryFn: ({ signal }) => apiClient.getExternalEvents(signal),
+  })
+
+  return (
+    <div className="flex flex-col gap-4">
+      <div className="flex flex-col gap-1">
+        <h1 className="text-2xl font-semibold">Facebook Events</h1>
+        <p className="text-muted-foreground text-sm">
+          Feature or cancel upcoming Facebook/Eventbrite events shown to
+          activists.
```
jakehobbs: change to say displayed on public events page here too

frontend-v2/src/app/(authed)/external-events/external-events-page.tsx
```diff
@@ -0,0 +1,40 @@
...
+        </p>
+      </div>
+
+      {isLoading ? (
+        <div className="flex items-center gap-2 text-sm text-muted-foreground">
+          <Loader2 className="h-4 w-4 animate-spin" />
+          Loading events...
+        </div>
+      ) : isError || !events ? (
```
jakehobbs: should error state be separated from empty state?

frontend-v2/src/app/(authed)/external-events/page.tsx
```diff
@@ -0,0 +1,36 @@
...
+export default async function ExternalEventsListPage() {
+  const session = await getCachedSession()
+  if (!session.user?.Roles.includes('admin')) {
+    forbidden()
+  }
+
+  const apiClient = new ApiClient(await getCookies())
+  const queryClient = new QueryClient()
+
+  // prefetchQuery, not fetchQuery + redirectForHttpError: the endpoint isn't
```
jakehobbs, line 22: actual good comment 👍

frontend-v2/src/lib/api.ts
```diff
@@ -293,6 +296,25 @@ export interface EventListParams {
   event_type: EventType
 }
 
+// Listing events for this page ID returns merged events from all SF Bay pages.
+const SF_BAY_FACEBOOK_PAGE_ID = '1377014279263790'
+
+// Go's ExternalEvent struct has no `json` tags, so fields serialize as PascalCase.
```
jakehobbs: this comment is not needed

frontend-v2/src/lib/api.ts
```diff
@@ -293,6 +296,25 @@ export interface EventListParams {
   event_type: EventType
 }
 
+// Listing events for this page ID returns merged events from all SF Bay pages.
+const SF_BAY_FACEBOOK_PAGE_ID = '1377014279263790'
```
jakehobbs, line 301: maybe comment should note that the event events admin page is only used for sf bay to justify this existing, parity w/ old vue page. the existing comment can be removed. keep it short.

frontend-v2/src/app/(authed)/external-events/external-events-table.tsx
jakehobbs: [Claude] Kept — this directive is the repo-wide convention for `useReactTable` pages: it appears verbatim in `users/user-table.tsx`, `events/event-list-table.tsx`, `activists/activists-table.tsx`, and `intl/organizers/organizers-page.tsx`. I verified that removing it makes eslint emit a `react-hooks/incompatible-library` warning, and CI runs `pnpm lint --max-warnings=0`, so lint would fail without it.

frontend-v2/src/app/(authed)/external-events/cancel-event-dialog.tsx
jakehobbs: [Claude] Kept `open` — it is the established pattern here: `hide-activist-dialog.tsx` and `merge-activist-dialog.tsx` both take `open`/`onOpenChange` props, mirroring the Radix Dialog API surfaced by `components/ui/dialog.tsx`.

frontend-v2/src/app/(authed)/external-events/cancel-event-dialog.tsx
jakehobbs: [Claude] Kept — this is an exact existing pattern: `hide-activist-dialog.tsx:48` and `merge-activist-dialog.tsx:61` both define `const handleOpenChange = (next: boolean)` with the same `isPending` guard.

frontend-v2/src/app/(authed)/external-events/cancel-event-dialog.tsx
jakehobbs: [Claude] Fixed in 243e93bf — the dialog now says it will no longer be displayed on the public events page.

frontend-v2/src/app/(authed)/external-events/external-events-page.tsx
jakehobbs: [Claude] Fixed in 243e93bf — the description now reads "Feature or cancel upcoming Facebook/Eventbrite events displayed on the public events page."

frontend-v2/src/app/(authed)/external-events/external-events-page.tsx
jakehobbs: [Claude] Fixed in 243e93bf. The endpoint always returns an array on success (a null response is normalized to `[]` in the zod schema), so after loading, `events` is undefined only when the fetch failed. The render logic now reads unambiguously: loading → spinner; data present → table (zero events shows the table's "No upcoming events found." empty state); otherwise → "Failed to load events" error notice. Error and empty are fully separate states.

frontend-v2/src/app/(authed)/external-events/page.tsx
jakehobbs: [Claude] Thanks! No change needed here.

frontend-v2/src/lib/api.ts
jakehobbs: [Claude] Fixed in 243e93bf — comment removed.

frontend-v2/src/lib/api.ts
jakehobbs: [Claude] Fixed in 243e93bf — replaced with a short note that the external events admin page is only used for SF Bay (parity with the legacy Vue page), which is why the page ID is hardcoded.

frontend-v2/src/app/(authed)/external-events/cancel-event-dialog.tsx
```diff
@@ -0,0 +1,80 @@
...
+          >
+            Keep event
+          </Button>
+          <Button
+            type="button"
+            variant="destructive"
+            onClick={() => mutation.mutate()}
+            disabled={mutation.isPending}
+          >
+            {mutation.isPending ? 'Cancelling...' : 'Cancel event'}
```
alexsapps, line 74:
```suggestion
            {mutation.isPending ? 'Deleting...' : 'Delete event'}
```

"Cancel event" button could be confused for "Cancel this delete operation". how would you feel about the buttons saying "Delete event" and "Keep event"

currently:
Image

alexsapps (approved):

PR #463, opened 2026-07-19
Port apply form to React

jakehobbs:
## Summary
- Ports the public "Apply" form (chapter member / organizer application) from `frontend/FormApply.vue` to Next.js at `frontend-v2/src/app/apply/` (URL `/v2/apply`).
- Public, unauthenticated page: lives outside `(authed)`, follows the `login` page pattern (no navbar, no auth provider).
- Submits to the existing, unmodified Go endpoint `POST /apply` (`ApplicationFormHandler` in `server/src/main.go`, unauthenticated, no CSRF middleware) with the same payload shape the Vue form sends: `{ name, firstName, lastName, pronouns, email, address, city, zip, phone, birthday, referral, language, accessibility, applicationType }`. Added a typed `submitApplicationForm` method + `ApplicationFormPayload` interface to `ApiClient` in `frontend-v2/src/lib/api.ts`.
- No backend changes, no changes to `frontend/`.

## Feature parity
- Kept all copy/text content, form fields, and required-ness identical.
- Same step flow: local-area gate -> chapter-member info copy + "Apply now" -> application fields -> thank-you message.
- Same validation order/messages as `submitForm()` in the Vue page: required fields block submission first (native HTML5 `required`, replacing Vue's `checkHtml5Validity()` refs check), then the same sequential toast checks for code-of-conduct, mission & values, consent, and application type (text preserved verbatim, including the original's minor grammar).
- "No" (not local) still does a full-page redirect to the legacy, still-Vue-served `/international` page, since that page hasn't been ported yet.

## Intentional differences
- Styling ported from Bulma/Buefy to Tailwind + the shared `frontend-v2/src/components/ui/` components, per frontend-v2 conventions.
- The three "I agree" toggles (mission/conduct/consent) and the organizer yes/no choice are custom-checked in JS (toast errors) rather than via native HTML5 required semantics, matching the Vue page's actual behavior (those weren't in its `refsToValidate` list either).
- Trimming: the legacy form used `v-model.trim`, which trims on every keystroke and can clip the space between words while typing a multi-word address/city (a pre-existing Vue quirk). The React version trims values once, at submit time, instead — same end result without the mid-typing bug.
- Zip code input uses `type="text" inputMode="numeric"` instead of `type="number"`, avoiding leading-zero issues with 5-digit zip codes.

## Test plan
- [x] `pnpm exec tsc --noEmit` — passes (pre-existing, unrelated `baseUrl` deprecation warning present on `main` too)
- [x] `pnpm lint` — passes
- [x] `pnpm build` — passes; `/apply` prerenders as a static page
- [x] Ran `pnpm start` and curled `http://localhost:3411/v2/apply` — 200 response, correct `<title>Join DxE SF Bay</title>`, correct heading and form gate content rendered


frontend-v2/src/app/apply/application-fields.tsx
```diff
@@ -0,0 +1,399 @@
...
+      address: address.trim(),
+      city: city.trim(),
+      zip: zip.trim(),
+      phone: phone.trim(),
+      birthday: birthday.trim(),
+      referral: referral.trim(),
+      language: language.trim(),
+      accessibility: accessibility.trim(),
+      applicationType,
+    })
```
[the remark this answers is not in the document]
jakehobbs: [Claude] Fixed in 5175ff09: submit now validates the trimmed value for every required field and blocks submission with a per-field error toast if a field is whitespace-only, instead of silently trimming and submitting a blank value.

frontend-v2/src/app/apply/application-fields.tsx
```diff
@@ -0,0 +1,399 @@
...
+      <p className="text-sm">{children}</p>
+      <label
+        htmlFor={id}
+        className="flex w-fit cursor-pointer items-center gap-2 rounded-md border p-2 text-sm"
+      >
+        <Checkbox
+          id={id}
+          checked={checked}
+          onCheckedChange={(value) => onCheckedChange(Boolean(value))}
+        />
```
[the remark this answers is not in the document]
jakehobbs: [Claude] Fixed in 5175ff09: the agreement statement paragraph now has an id referenced by `aria-describedby` on the checkbox, so screen readers announce the full statement alongside the "Yes, I agree" label.

frontend-v2/src/app/apply/apply-form.tsx
```diff
@@ -0,0 +1,80 @@
...
+    mutationFn: (payload: ApplicationFormPayload) =>
+      apiClient.submitApplicationForm(payload),
+    onSuccess: () => {
+      toast.success('Submitted!')
+      setSubmitSuccess(true)
+    },
+    onError: () => {
+      toast.error(ERROR_MESSAGE)
+    },
+  })
```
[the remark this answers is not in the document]
jakehobbs: [Claude] Acknowledged — this is a reasonable suggestion, but we're not adding it in this PR to keep the port reviewable in scope. Noted for a follow-up.

frontend-v2/src/app/apply/application-fields.tsx
```diff
@@ -0,0 +1,421 @@
...
+  const [accessibility, setAccessibility] = useState('')
+  const [applicationType, setApplicationType] = useState<ApplicationType>('')
+
+  function handleSubmit(e: FormEvent<HTMLFormElement>) {
+    e.preventDefault()
+
+    // Trim at submit time rather than on every keystroke (unlike the legacy
+    // Vue form's `v-model.trim`, which trims per-keystroke and can clip the
+    // space between words while typing a multi-word address/city).
+    const trimmed = {
```
jakehobbs: can't this be done in the zod schema?

frontend-v2/src/app/apply/application-fields.tsx
```diff
@@ -0,0 +1,421 @@
...
+    ]
+    for (const [field, label] of requiredFields) {
+      if (!trimmed[field]) {
+        toast.error(`${label} is required.`)
+        return
+      }
+    }
+
+    // Business-rule checks, same order as the legacy Vue submitForm().
+    if (!conduct) {
```
jakehobbs: seems like a lot of manual validation here. this should be handled within zod.

frontend-v2/src/app/apply/application-fields.tsx
jakehobbs: [Claude] Done in be656202 — the trimming and required-field checks now live in a zod schema (`z.string().trim().min(1, ...)` per required field), and the payload is built from the parsed (trimmed) output.

frontend-v2/src/app/apply/application-fields.tsx
jakehobbs: [Claude] Done in be656202 — all manual validation is replaced by a single `applicationSchema` zod object: required fields via `.trim().min(1)`, agreements via `z.literal(true)`, and applicationType via `z.enum`, each carrying the legacy error message. Submit does one `safeParse` and toasts the first issue; schema key order preserves the legacy validation order (required fields, then conduct -> mission -> consent -> applicationType).

frontend-v2/src/app/apply/application-fields.tsx
```diff
@@ -0,0 +1,418 @@
...
+  applicationType: z.enum(['organizer', 'chapter-member'], {
+    message:
+      'You must choose whether or not you are interested in becoming an organizer.',
+  }),
+  referral: z.string().trim(),
+  language: z.string().trim(),
+  accessibility: z.string().trim(),
+})
+
+type ApplicationType = z.infer<typeof applicationSchema>['applicationType'] | ''
```
jakehobbs: it feels a little weird to me to put empty string in the type

frontend-v2/src/app/apply/apply-form.tsx
```diff
@@ -0,0 +1,63 @@
...
+import { useState } from 'react'
+import { useMutation } from '@tanstack/react-query'
+import toast from 'react-hot-toast'
+import { apiClient, ApplicationFormPayload } from '@/lib/api'
+import { LocalCheck } from '@/app/apply/local-check'
+import { ChapterMemberInfo } from '@/app/apply/chapter-member-info'
+import { ApplicationFields } from '@/app/apply/application-fields'
+import { ThankYou } from '@/app/apply/thank-you'
+
+const ERROR_MESSAGE =
```
jakehobbs: no need for const if used in 1 place

frontend-v2/src/app/apply/application-fields.tsx
jakehobbs: [Claude] Fixed in a3ef4e4e — the `| ''` sentinel is gone. `ApplicationType` is now just the schema enum ('organizer' | 'chapter-member'), and the unselected state is modeled as `ApplicationType | null` (default `null`) on the tanstack-form field value; the schema enum rejects `null` with the legacy error message.

frontend-v2/src/app/apply/apply-form.tsx
jakehobbs: [Claude] Fixed in a3ef4e4e — ERROR_MESSAGE was only used once, so it is now inlined at the toast.error call site in the mutation onError handler.

alexsapps (approved):

PR #464, opened 2026-07-19
Port working groups page to React

jakehobbs:
> [!NOTE]
> Stacked on #468 (`jake/react-shared-tag-input`), which introduces the shared `TagInput` component this page uses for the Point Person / Members / Non-members pickers. Merge #468 first; this PR's base is set to that branch so the diff stays clean.

## Summary

Ports the legacy Vue **Working Groups** admin page (`frontend/WorkingGroupList.vue`, served at `/list_working_groups`) to the Next.js app at `/v2/working-groups`. The Vue page is untouched and still works as a "Legacy" nav entry.

## Endpoints used (no Go changes)

- `GET working_group/list` — `WorkingGroupListHandler`
- `POST working_group/save` — `WorkingGroupSaveHandler` (create when no `id`, update otherwise)
- `POST working_group/delete` — `WorkingGroupDeleteHandler`
- `GET activist_names/get_organizers` — for the Point Person / Members pickers
- `GET activist_names/get` — for the Non-members on Mailing List picker

All four `working_group/*` endpoints are gated server-side by `apiSFBayOrganizerAuthMiddleware` (admin, or organizer + SF Bay chapter), same as the legacy page's `authSFBayOrganizerMiddleware`. The new React page doesn't duplicate that role check — it relies on the existing `redirectForHttpError` SSR pattern (same as `users/page.tsx`): if a non-authorized user's prefetch throws an `HTTPStatusError(403)`, Next's `forbidden()` page is shown.

- "Groups" → "Working Groups" now points at `/v2/working-groups` with `page: WorkingGroupList_beta`.
- Added the legacy entry to the end of the "Legacy" dropdown (`href: /list_working_groups`, `page: WorkingGroupList`, `roleRequired: [organizer]`, `visibleForNonSFBay: false`), so the old page stays reachable.

## Feature parity

- List with sortable Name/Email columns, Point Person column, and a "Show/Hide members" toggle switching between a "Total Members" count column and a full "Members" list column (desktop table + mobile card view, matching other list pages like `users/`).
- Create/Edit via a dialog: name, email, description, meeting day & time, meeting location, visible switch, and three chip-based pickers (Point Person — max 1, Members, Non-members on Mailing List) with type-ahead search, mirroring the Buefy `b-taginput` behavior (existing names only, no free-form entries).
- Delete via a confirmation dialog with the same warning as the legacy page ("remove all members before deleting"); the backend still rejects deletion of a working group that has members, surfaced via a toast.
- Preserved a couple of small existing quirks from the Vue page rather than "fixing" them, to keep behavior identical:
  - The "Total Members" count includes the point person but excludes non-mailing-list members; the "Members" list (when expanded) includes non-mailing-list members. This mismatch exists in the legacy page too.
  - The `coords` field (present in the Go model, never exposed in the legacy UI) is round-tripped unmodified on save so it isn't silently cleared.

## Intentional differences

- Create/edit uses a shared `Dialog` component instead of Buefy modals — same UX, matches this app's existing conventions (`Dialog` used elsewhere; no other in-place-edit page in `frontend-v2` uses a Buefy-style full-screen modal).
- The chip/type-ahead pickers use the shared `TagInput` component from #468 (also used by the circles page) rather than the heavier `events/` attendee-autocomplete machinery (IndexedDB-backed activist registry), since working groups only need simple, bounded name lists (organizers, activists) with no offline caching requirement. Its suggestion filtering is a case-insensitive prefix match, matching the legacy Vue `getFilteredOrganizers`/`getFilteredActivists` behavior.

## Test plan

- [x] `pnpm exec tsc --noEmit` — fails only on a pre-existing, unrelated `tsconfig.json` `baseUrl` deprecation warning from the `typescript@6.0.3` devDependency; confirmed via `git stash` that this reproduces identically on the unmodified branch.
- [x] `pnpm lint` — clean
- [x] `pnpm build` — succeeds; Next's own internal TypeScript check (part of `next build`) passes cleanly and `/working-groups` is listed in the route output
- [x] `pnpx prettier --write` run on every changed/created file
- [ ] Manual click-through in a running dev environment (list, create, edit, delete, member visibility toggle, non-SF-Bay/non-organizer 403 behavior)


frontend-v2/src/lib/api.ts
```diff
@@ -416,6 +471,13 @@ export class ApiClient {
     return ActivistNamesResp.parse(resp)
   }
 
+  getOrganizerNames = async (signal?: AbortSignal) => {
+    const resp = await this.client
+      .get(API_PATH.ACTIVIST_NAMES_GET_ORGANIZERS, { signal })
+      .json()
+    return ActivistNamesResp.parse(resp)
+  }
```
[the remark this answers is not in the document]
jakehobbs: [Claude] Fixed in 18e13661: `getOrganizerNames` and `getActivistNames` now wrap ky's `HTTPError` via `handleKyError`, so an SSR 403 from either endpoint produces an `HTTPStatusError` that `redirectForHttpError` can act on, instead of an unwrapped error.

frontend-v2/src/app/(authed)/working-groups/person-multi-select.tsx
```diff
@@ -0,0 +1,126 @@
...
+            <input
+              className="min-w-[8rem] flex-1 bg-transparent text-sm outline-none placeholder:text-muted-foreground"
+              value={query}
+              placeholder={value.length === 0 ? placeholder : ''}
+              onChange={(e) => {
+                setQuery(e.target.value)
+                setIsOpen(true)
+              }}
+              onFocus={() => setIsOpen(true)}
+              onBlur={() => setIsOpen(false)}
```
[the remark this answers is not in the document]
jakehobbs: [Claude] Obsolete — `person-multi-select.tsx` was deleted from this PR when it adopted the shared `TagInput` component introduced in #468. The equivalent labeling/ARIA concerns were addressed in that shared component (commit daced44e on #468).

frontend-v2/src/app/(authed)/working-groups/delete-working-group-dialog.tsx
```diff
@@ -0,0 +1,77 @@
...
+        </DialogHeader>
+        <p className="text-sm">
+          Are you sure you want to delete <strong>{workingGroup.name}</strong>?
+        </p>
+        <div className="flex items-start gap-2 rounded-md border border-yellow-300 bg-yellow-50 p-3 text-sm text-yellow-900">
+          <AlertTriangle className="mt-0.5 h-4 w-4 shrink-0" />
+          <span>
+            Before deleting a working group, be sure to remove all members of
+            that group.
+          </span>
```
jakehobbs, line 60: can we enforce this? at least best-effort client side?

frontend-v2/src/app/(authed)/working-groups/working-group-form-dialog.tsx
```diff
@@ -0,0 +1,330 @@
...
+    defaultValues: initialValues,
+    onSubmit: async ({ value }) => {
+      const name = value.name.trim()
+      if (!name) {
+        toast.error('Working group name must not be blank')
+        return
+      }
+      const email = value.email.trim()
+      if (!email.includes('@')) {
+        toast.error(`Working group email must contain @: ${email}`)
```
jakehobbs: this should be handled by the zod schema

frontend-v2/src/app/(authed)/working-groups/working-group-form-dialog.tsx
```diff
@@ -0,0 +1,330 @@
...
+        })
+      }
+      for (const m of value.nonMembers) {
+        if (m === pointPersonName) continue
+        members.push({
+          name: m,
+          point_person: false,
+          non_member_on_mailing_list: true,
+        })
+      }
```
jakehobbs: this feels very iterative. can we just create the members array be spreading from a map of each source array instead to follow more common react patterns?

frontend-v2/src/app/(authed)/working-groups/working-groups-page.tsx
```diff
@@ -0,0 +1,127 @@
...
+          </Button>
+        </div>
+      </div>
+
+      {isLoading ? (
+        <div className="flex items-center gap-2 text-muted-foreground text-sm">
+          <Loader2 className="h-4 w-4 animate-spin" />
+          Loading working groups...
+        </div>
+      ) : isError || !workingGroups ? (
```
jakehobbs: separate empty state from error state?

frontend-v2/src/app/(authed)/working-groups/working-groups-page.tsx
```diff
@@ -0,0 +1,127 @@
...
+          <WorkingGroupTable
+            workingGroups={workingGroups}
+            membersVisible={membersVisible}
+            onEdit={setEditingTarget}
+            onDelete={setDeletingTarget}
+          />
+        </>
+      )}
+
+      {editingTarget !== null && (
```
jakehobbs: != null

frontend-v2/src/app/(authed)/working-groups/delete-working-group-dialog.tsx
jakehobbs: [Claude] Done in 3b340e8f. Checked the backend first: DeleteWorkingGroup (model/working_groups.go) already rejects deletion when any working_group_members rows exist. The dialog now mirrors that best-effort client-side — when the group still has members (including non-members on the mailing list), the Delete button is disabled and the warning explains that they must be removed first. The server check still catches stale client data.

frontend-v2/src/app/(authed)/working-groups/working-group-form-dialog.tsx
jakehobbs: [Claude] Done in 3b340e8f. Name/email validation now lives in a zod schema (mirroring circle-form-dialog.tsx): name is trim + min(1), email is trim + refine(includes('@')) with a dynamic message, and onSubmit safeParses and toasts the first issue. Message text is unchanged.

frontend-v2/src/app/(authed)/working-groups/working-group-form-dialog.tsx
jakehobbs: [Claude] Done in 3b340e8f. The members array is now built by spreading three sources: the optional point-person entry, then members and non-members mapped with a shared notPointPerson filter. Ordering (point person first), dedup rules, and the legacy quirk where a name in both lists appears twice are all preserved — the payload is identical to the previous imperative version for every input.

frontend-v2/src/app/(authed)/working-groups/working-groups-page.tsx
jakehobbs: [Claude] Done in 3b340e8f. The conflated isError || !workingGroups condition is gone: loading shows the spinner, a fetch failure shows the error notice, and success renders the table, whose own "No working groups found." empty state handles an empty array — same shape as the fix in #462.

frontend-v2/src/app/(authed)/working-groups/working-groups-page.tsx
jakehobbs: [Claude] Done in 3b340e8f — switched to editingTarget != null. Checked the rest of the file for the same pattern; that was the only strict-null comparison.

alexsapps (approved):

PR #465, opened 2026-07-19
Port interest form to React

jakehobbs:
## Summary
- Ports the public "Interest" sign-up form (`frontend/FormInterest.vue`, served at the legacy Vue route `/interest`) to Next.js at `frontend-v2/src/app/interest/`, reachable at `/v2/interest` (outside the `(authed)` route group — no navbar, no auth provider, matching the `/login` page pattern).
- No backend changes. No changes to `frontend/` (the Vue page keeps working).

## Endpoint + payload
- Submits via `POST /interest` (existing, unauthenticated `InterestFormHandler` Go route — registered outside any CSRF/auth middleware, so no CSRF token is sent).
- New typed `ApiClient.submitInterestForm()` method in `frontend-v2/src/lib/api.ts` posts the same payload shape the Vue form used, matching `model.InterestFormData`'s JSON tags exactly:
  ```ts
  { chapterId, form, name, email, zip, phone, referralFriends, referralApply, referralOutlet, interests }
  ```
- Response handling matches the existing "legacy handler reports errors via 200 + `{status:"error"}`" pattern already used elsewhere in `ApiClient` (`throwIfApiError`).

## Query params handled
Reads the exact same URL query params the Vue form parses in its `created()` hook — the same params generated by the existing "Interest Form Generator" at `/interest/generator` (unchanged, still points at `/interest`):
- `chapterId` (falls back to `SF_BAY_CHAPTER_ID` if missing/invalid)
- `name` (form name; defaults to `"Interest Form"`)
- `title` (defaults based on chapter: "DxE SF Bay - Get Involved" vs "Direct Action Everywhere - Get Involved")
- `description`
- `showInterests` (defaults to **true**, hidden only if explicitly `"false"`)
- `showReferralFriends` / `showReferralApply` / `showReferralOutlet` (default **false**, shown only if explicitly `"true"`)
- `referralApply` (pre-fills the "who encouraged you" field unless literally `"null"`)

## Feature parity notes / intentional differences
- All copy, fields (first/last name, email, phone, zip — all required, same `maxlength`s), the 5 activism-interest checkboxes, the referral-friends/referral-apply text fields, the referral-outlet radio group (4 options), and both success states (generic "Thank you for your submission!" vs. the "Check-in" form's personalized "Thank you, `<first> <last>`!" + "Submit another form" button) are ported with full parity.
- **Validation deviation**: the legacy form used native browser validation bubbles (`checkHtml5Validity()`), so there was no legacy message text to preserve. The React form is `noValidate` and validates via a zod schema (`@tanstack/react-form` onSubmit validator) with a first-issue error toast plus inline field errors — message style consistent with the sibling apply/international forms (e.g. "First Name is required.", "Enter a valid email."). Submission goes through a react-query `useMutation` wrapping `apiClient.submitInterestForm`.
- Added a new `RadioGroup`/`RadioGroupItem` UI primitive (`frontend-v2/src/components/ui/radio-group.tsx`, using the newly-added `@radix-ui/react-radio-group` dep) since no radio component existed yet in `components/ui/` — built in the same style as the existing `Checkbox` component.
- **Intentionally not ported**: the Vue form's `circleOptions`/`circleInterests` state (fetched from `GET /circles` on mount) is dead code — it's fetched but never bound to any rendered checkbox in the Vue template, so `circleInterests` is always `[]`. The "Circle Interest" form name branch therefore always submitted an empty `interests` string in the legacy app; this port reproduces that exact behavior directly (`interests: formName === 'Circle Interest' ? '' : activismInterests.join(', ')`) without fetching `/circles` at all, since doing so would have no observable effect.
- Icons that decorated Buefy inputs (alphabetical/email/phone/city/account icons) were dropped since `frontend-v2`'s `Input` component has no icon slot — a minor cosmetic difference, not a functional one.

## Test plan
- [x] `pnpm exec tsc --noEmit` — passes (one pre-existing, unrelated `tsconfig.json` `baseUrl` deprecation warning reproduces identically with these changes stashed out, so it's environmental, not introduced by this PR)
- [x] `pnpm lint` — clean
- [x] `pnpm build` — succeeds; `/interest` builds as a separate static route from `/interest/generator` (route groups don't collide)
- [x] `pnpx prettier --write` run on every changed/added file
- [x] Ran `pnpm dev` and drove `/v2/interest` end-to-end with Playwright (system Chrome):
  - Default (no query params): correct title, all 5 required fields, activism-interest checkboxes visible, referral sections hidden
  - Custom query params (`chapterId`, `name=Check-in`, `title`, `description`, `showInterests=false`, all three `showReferral*=true`, `referralApply=Jane`): custom title/description rendered, activism interests hidden, referral-friends/apply/outlet all shown, `referralApply` field pre-filled with "Jane"
  - Submitting with empty required fields is blocked by zod validation (no network call made; first-issue toast "First Name is required." plus inline field errors; invalid email shows "Enter a valid email.")
  - Submitting valid data POSTs to `/interest` with the expected JSON body
  - Mocked a successful `/interest` response: "Check-in" form name shows the personalized thank-you + "Submit another form" button; default form name shows the generic thank-you message


frontend-v2/src/lib/api.ts
```diff
@@ -297,6 +298,22 @@ const SuccessResp = z.object({
   status: z.literal('success'),
 })
 
+// Payload for the public (unauthenticated) "Interest" sign-up form, submitted
+// from frontend-v2/src/app/interest. Field names/shape must match
+// model.InterestFormData in server/src/model/forms.go exactly.
```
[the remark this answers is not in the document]
jakehobbs: [Claude] Fixed in 0e3c7b61: the doc comment is reworded to note the Go struct's extra `submitted_via_signup_service` field that this form never sends, rather than claiming an exact match.

frontend-v2/src/app/interest/interest-form.tsx
```diff
@@ -0,0 +1,332 @@
...
+  // Form options are derived once from the URL query params, mirroring the
+  // `created()` hook in frontend/FormInterest.vue.
+  const formOptions = useMemo(() => {
+    const chapterIdStr = searchParams.get('chapterId')
+    const parsedChapterId =
+      chapterIdStr != null ? parseInt(chapterIdStr, 10) : NaN
+    const chapterId = !Number.isNaN(parsedChapterId)
+      ? parsedChapterId
+      : SF_BAY_CHAPTER_ID
+    const formName = searchParams.get('name') || 'Interest Form'
```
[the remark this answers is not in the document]
jakehobbs: [Claude] Acknowledged — reasonable suggestion, but not added in this PR to keep scope focused on the port. Noted for a follow-up.

frontend-v2/src/app/interest/interest-form.tsx
```diff
@@ -0,0 +1,331 @@
...
+import toast from 'react-hot-toast'
+import { Input } from '@/components/ui/input'
+import { Label } from '@/components/ui/label'
+import { Checkbox } from '@/components/ui/checkbox'
+import { RadioGroup, RadioGroupItem } from '@/components/ui/radio-group'
+import { Button } from '@/components/ui/button'
+import { apiClient } from '@/lib/api'
+import { SF_BAY_CHAPTER_ID } from '@/lib/constants'
+
+const ERROR_MESSAGE =
```
jakehobbs: no need for this message to be a const since only used in 1 place

frontend-v2/src/app/interest/interest-form.tsx
```diff
@@ -0,0 +1,331 @@
...
+  { value: 'Email', label: 'Email' },
+  { value: 'Meetup', label: 'Saturday morning meetup' },
+  { value: 'In-person Invite', label: 'Someone invited me in person' },
+] as const
+
+export function InterestForm() {
+  const searchParams = useSearchParams()
+
+  // Mirrors the query-param parsing in FormInterest.vue's created() hook.
+  const formOptions = useMemo(() => {
```
jakehobbs: should we use nuqs for this? or does it add too big of a refactor/complexity?

frontend-v2/src/app/interest/interest-form.tsx
```diff
@@ -0,0 +1,331 @@
...
+    const showInterests = searchParams.get('showInterests') !== 'false'
+    const showReferralFriends =
+      searchParams.get('showReferralFriends') === 'true'
+    const showReferralApply = searchParams.get('showReferralApply') === 'true'
+    const showReferralOutlet = searchParams.get('showReferralOutlet') === 'true'
+    const referralApplyParam = searchParams.get('referralApply') || ''
+    const initialReferralApply =
+      referralApplyParam !== 'null' ? referralApplyParam : ''
+
+    return {
```
jakehobbs: maybe we should extract a helper to go from query params to the "show" vars, etc. that is outside the component?

frontend-v2/src/app/interest/interest-form.tsx
jakehobbs: [Claude] Inlined in af3b4dfe — the constant was single-use, so the message now lives directly in the mutation's onError toast.

frontend-v2/src/app/interest/interest-form.tsx
jakehobbs: [Claude] I looked into it and would recommend against nuqs here. Its value is two-way URL⇄state sync — typed `useQueryState` setters that write filter/sort state back into the URL, like on the activists page. This form only reads the params once on load to configure itself and never writes back, so nuqs would just add a parser layer over what is a one-shot `URLSearchParams` read with legacy-specific defaulting quirks (e.g. `showInterests` defaults true while the other flags default false, and the title default depends on the pre-fallback chapterId). Instead I extracted that parsing into a pure `parseInterestFormOptions()` helper in `form-options.ts` (af3b4dfe, per the comment below), which gets the testability/organization benefit without a new dependency pattern. Happy to switch to nuqs if you'd prefer consistency over minimalism here.

frontend-v2/src/app/interest/interest-form.tsx
jakehobbs: [Claude] Done in af3b4dfe — extracted `parseInterestFormOptions(searchParams: URLSearchParams)` as a pure function in `frontend-v2/src/app/interest/form-options.ts`, keeping the legacy-parity comments (title defaulting from the pre-fallback chapterId, the show-flag default asymmetry) on the helper. The component now just calls it in a `useMemo`. Same commit also converts the form to @tanstack/react-form + a react-query `useMutation` per the other feedback.

frontend-v2/src/lib/api.ts
```diff
@@ -343,6 +344,21 @@ const SuccessResp = z.object({
   status: z.literal('success'),
 })
 
+// Public interest form payload. Keys match model.InterestFormData's JSON
+// tags (server/src/model/forms.go) for the fields this form submits.
```
alexsapps, line 359:
```suggestion
// tags for the fields this form submits.
```

frontend-v2/src/lib/api.ts
```diff
@@ -796,6 +812,19 @@ export class ApiClient {
       return this.handleKyError(err)
     }
   }
+
+  // Public endpoint; no CSRF token (route is outside the CSRF middleware group).
```
alexsapps, line 882:
```suggestion
```

frontend-v2/src/app/interest/page.tsx
```diff
@@ -0,0 +1,15 @@
+import { Suspense } from 'react'
+import { ContentWrapper } from '@/app/content-wrapper'
+import { InterestForm } from './interest-form'
+
+// Public sign-up form ported from frontend/FormInterest.vue. Query params
+// (built by the authed /interest/generator page) control which fields show.
```
alexsapps, line 6:
```suggestion
// Public interest / check-in forms. Query params (built by the
// /interest/generator page) control which fields show.
```

frontend-v2/src/app/interest/form-options.ts
```diff
@@ -0,0 +1,54 @@
...
+  formTitle: string
+  formDescription: string
+  showInterests: boolean
+  showReferralFriends: boolean
+  showReferralApply: boolean
+  showReferralOutlet: boolean
+  initialReferralApply: string
+}
+
+// Mirrors the query-param parsing in FormInterest.vue's created() hook.
```
alexsapps, line 15:
```suggestion
```

frontend-v2/src/app/interest/form-options.ts
```diff
@@ -0,0 +1,54 @@
...
+  const formName = searchParams.get('name') || 'Interest Form'
+  // Vue parity: the default title checks the raw parsed chapterId (NaN
+  // when missing/invalid), not the SF-Bay-defaulted value.
+  const formTitle =
+    searchParams.get('title') ||
+    (parsedChapterId === SF_BAY_CHAPTER_ID
+      ? 'DxE SF Bay - Get Involved'
+      : 'Direct Action Everywhere - Get Involved')
+  const formDescription = searchParams.get('description') || ''
+  // Vue parity: showInterests defaults true; other show* flags default false.
```
alexsapps, line 34:
```suggestion
// showInterests defaults true to preserve behavior of links generated
// before visibility of this section was configurable 
```

frontend-v2/src/app/interest/form-options.ts
```diff
@@ -0,0 +1,54 @@
...
+): InterestFormOptions {
+  const chapterIdStr = searchParams.get('chapterId')
+  const parsedChapterId =
+    chapterIdStr != null ? parseInt(chapterIdStr, 10) : NaN
+  const chapterId = !Number.isNaN(parsedChapterId)
+    ? parsedChapterId
+    : SF_BAY_CHAPTER_ID
+  const formName = searchParams.get('name') || 'Interest Form'
+  // Vue parity: the default title checks the raw parsed chapterId (NaN
+  // when missing/invalid), not the SF-Bay-defaulted value.
```
alexsapps, line 27: i don't understand this comment but probably just delete

frontend-v2/src/app/interest/interest-form.tsx
```diff
@@ -0,0 +1,391 @@
...
+
+const REFERRAL_OUTLET_OPTIONS = [
+  { value: 'Social Media', label: 'Social Media' },
+  { value: 'Email', label: 'Email' },
+  { value: 'Meetup', label: 'Saturday morning meetup' },
+  { value: 'In-person Invite', label: 'Someone invited me in person' },
+] as const
+
+// Unlike the legacy form (native browser validation bubbles), validation is
+// via zod with the form set to noValidate; `required` attrs are a11y-only.
```
alexsapps, line 57:
```suggestion
```

alexsapps (approved):

frontend-v2/src/lib/api.ts
jakehobbs: Applied — dropped the file path from the comment.

frontend-v2/src/lib/api.ts
jakehobbs: Applied — comment removed.

frontend-v2/src/app/interest/page.tsx
jakehobbs: Applied.

frontend-v2/src/app/interest/form-options.ts
jakehobbs: Applied — comment removed.

frontend-v2/src/app/interest/form-options.ts
jakehobbs: Applied.

frontend-v2/src/app/interest/form-options.ts
jakehobbs: Deleted. It was noting that the SF Bay title check uses the raw parsed `chapterId` (`NaN` when the param is missing/invalid) rather than the SF-Bay-defaulted value, matching the Vue form — but the code reads fine without it.

frontend-v2/src/app/interest/interest-form.tsx
jakehobbs: Applied — comment removed.

PR #466, opened 2026-07-19
Port circles page to React

jakehobbs:
> [!NOTE]
> **Stacked on #468** (`jake/react-shared-tag-input`), which provides the shared `TagInput` component this page uses for the host/members fields. Merge #468 first.

## Summary

Ports `frontend/CirclesList.vue` (~510 lines) to Next.js at a single new route, `/v2/circles`, with the "Interest Circles" vs "Geo-Circles" mode selected by a `?type=interest|geo` URL query param (default `interest`) managed via `nuqs` — mirroring how the legacy Vue component is mounted twice with a different `title` prop from the same Go template (`server/templates/circles_list.html`), driven by `ListCirclesHandler`/`ListGeoCirclesHandler`.

The legacy Vue page (`frontend/CirclesList.vue`) and its routes (`/list_circles`, `/list_geocircles`) are untouched and still work.

## Endpoints used (all pre-existing, no backend changes)

- `POST /circle/list` — fetches all circle groups (both types in one flat list, same as the Vue page); the new client-side filters by `type` param just like the legacy component's `this.title === 'GeoCirclesList' ? c.type === 'geo-circle' : c.type === 'circle'` check.
- `POST /circle/save` — create (`id: 0`) or update (`id: <existing>`) a circle group.
- `POST /circle/delete` — delete by `circle_id`; the backend refuses if the circle still has members, matching the legacy delete-modal warning.
- `GET /activist_names/get_chaptermembers` — chapter-member/organizer names for the host/members autocomplete (this is a different, more restrictive endpoint than `activist_names/get`, matching exactly what the Vue page calls).

Added typed `ApiClient` methods (`getCircles`, `saveCircle`, `deleteCircle`, `getChapterMemberActivistNames`) and `API_PATH` constants (`CIRCLE_LIST`, `CIRCLE_SAVE`, `CIRCLE_DELETE`, `ACTIVIST_NAMES_CHAPTER_MEMBERS`) plus zod schemas (`CircleGroupSchema`, `CircleMemberSchema`) in `frontend-v2/src/lib/api.ts`.

## Access gating

The `/circle/*` endpoints are backed by `apiSFBayOrganizerAuthMiddleware` in Go (admin always allowed regardless of chapter; organizer only if the user's chapter is SF Bay). Rather than duplicating that role logic client-side, `page.tsx` wraps the initial `circle/list`/`activist_names/get_chaptermembers` prefetch in `redirectForHttpError` (the `users/page.tsx` pattern), so the backend's real 403 surfaces as Next's forbidden UI. This correctly preserves the backend's unconditional admin bypass for non-SF-Bay admins.

## Nav changes (`shared/nav.json` only)

- "Groups" dropdown: "Geo-Circles" → `/v2/circles?type=geo`, page `CirclesList_beta`; "Interest Circles" → `/v2/circles?type=interest`, page `CirclesList_beta`.
- Appended "Geo-Circles" and "Interest Circles" (pointing at the original `/list_geocircles`/`/list_circles` legacy routes) to the end of the "Legacy" dropdown, each with `roleRequired: ["organizer"]`, `visibleForNonSFBay: false`.

## Feature parity / intentional differences

- Full parity: list with sortable Name/Last Event columns, create/edit/delete modals, host (single) + members (geo-only) autocomplete, "Description"/"Description or Notes" label swap, meeting time/location fields (interest-only), coordinates field (disabled for geo, since the backend recomputes it from member locations), "Visible to public" toggle, geo-mode "Show/Hide members" toggle with a "Total Members" vs per-member list column, and the last-meeting freshness color coding (red/yellow/green based on days since, using `date-fns` instead of the Vue page's `dayjs`).
- The host/members fields use the shared `TagInput` component (chip list + `Popover`-based autocomplete filtered by "starts with", matching `getFilteredActivists`) from #468 in place of Buefy's `b-taginput`.
- The "Visible to public" toggle uses the existing `Checkbox` primitive (no `Switch` component exists in this codebase yet) instead of a switch control.
- Table sorting is implemented with local component state (two sortable columns: Name, Last Event) rather than TanStack Table, since the existing TanStack Table usage in this codebase (e.g. `user-table.tsx`) is overkill for this dataset size.

## Test plan

- [x] `cd frontend-v2 && pnpm exec tsc --noEmit --ignoreDeprecations 6.0` — no errors from this change. (Note: without `--ignoreDeprecations 6.0`, `tsc` fails on `tsconfig.json`'s `baseUrl` option with `TS5101`; confirmed via `git stash -u` that this is pre-existing/environmental, present on `main` before this change, likely from the `typescript: ^6.0.3` dependency bump vs. the current `tsconfig.json`. Next's own build-time type checking, run below, is unaffected.)
- [x] `pnpm lint` — clean.
- [x] `pnpm build` — succeeds; `/circles` appears in the route list. (Confirmed the `$public/logo.png` module-resolution warning some tooling surfaces in `nav.tsx` is also pre-existing/unrelated, via the same `git stash -u` check.)
- [x] `pnpm dev` + `curl /v2/circles?type=geo` — page compiles and renders (200); the only runtime error is a missing `NEXT_PUBLIC_API_BASE_URL`/live Go backend in this standalone check, expected since no backend/DB was running.
- [ ] Manual QA against a running Go backend: create/edit/delete an interest circle and a geo-circle, verify host/member assignment, the members-visible toggle, and that non-SF-Bay organizers/non-organizers are forbidden from `/v2/circles`.


frontend-v2/src/lib/api.ts
```diff
@@ -297,6 +301,62 @@ const SuccessResp = z.object({
...
+    .array(CircleMemberSchema)
+    .nullable()
+    .transform((v) => v ?? []),
+  visible: z.boolean(),
+  description: z.string(),
+  meeting_time: z.string(),
+  meeting_location: z.string(),
+  coords: z.string(),
+  last_meeting: z.string(),
+})
```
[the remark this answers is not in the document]
jakehobbs: [Claude] Fixed in d931b466: `CircleGroupSchema.type` is now a `z.enum` over the circle type values instead of a plain string, matching the restriction already applied to `SaveCircleParams.type`.

frontend-v2/src/lib/api.ts
```diff
@@ -750,6 +810,64 @@ export class ApiClient {
...
+  getChapterMemberActivistNames = async (signal?: AbortSignal) => {
+    try {
+      const resp = await this.client
+        .get(API_PATH.ACTIVIST_NAMES_CHAPTER_MEMBERS, { signal })
+        .json()
+      return ActivistNamesResp.parse(resp).activist_names
+    } catch (err) {
+      return this.handleKyError(err)
+    }
+  }
```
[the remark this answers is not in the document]
jakehobbs: [Claude] Fixed in d931b466 for consistency with the client's other handlers. Note the current Go handler behind this endpoint can't actually emit a soft `{status:"error"}` (it panics to a 500 instead), so this wasn't a live bug, but the `throwIfApiError` call is now added as defense-in-depth.

frontend-v2/src/app/(authed)/circles/circles-page.tsx
```diff
@@ -0,0 +1,157 @@
...
+  } = useQuery({
+    queryKey: [API_PATH.CIRCLE_LIST],
+    queryFn: ({ signal }) => apiClient.getCircles(signal),
+  })
+
+  const { data: activistNames, isLoading: isActivistsLoading } = useQuery({
+    queryKey: [API_PATH.ACTIVIST_NAMES_CHAPTER_MEMBERS],
+    queryFn: ({ signal }) => apiClient.getChapterMemberActivistNames(signal),
+  })
+
```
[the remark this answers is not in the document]
jakehobbs: [Claude] Fixed in d931b466: a failed activist-names query now surfaces an inline error notice on the page instead of silently leaving the host/members autocomplete empty and non-functional.

frontend-v2/src/app/(authed)/circles/circle-form-dialog.tsx
```diff
@@ -0,0 +1,314 @@
...
+  circle,
+  activistNames,
+}: Props) {
+  const queryClient = useQueryClient()
+  const isGeo = mode === 'geo'
+  const isEditing = circle !== null
+
+  const mutation = useMutation({
+    mutationFn: (value: z.output<typeof circleFormSchema>) => {
+      const memberParams: SaveCircleMemberParams[] = []
```
jakehobbs: let's create this array by spreading and mapping instead of pushing and forEaching

frontend-v2/src/app/(authed)/circles/circle-table.tsx
```diff
@@ -0,0 +1,231 @@
...
+        ),
+      },
+      {
+        id: 'host',
+        header: 'Host',
+        accessorFn: (row) => findPointPerson(row.members)?.name ?? '',
+      },
+    ]
+
+    if (mode === 'interest') {
```
jakehobbs: pls just spread these cols into array above instead of pushing

frontend-v2/src/app/(authed)/circles/circles-page.tsx
```diff
@@ -0,0 +1,168 @@
...
+  const {
+    data: activistNames,
+    isLoading: isActivistsLoading,
+    isError: isActivistsError,
+  } = useQuery({
+    queryKey: [API_PATH.ACTIVIST_NAMES_CHAPTER_MEMBERS],
+    queryFn: ({ signal }) => apiClient.getChapterMemberActivistNames(signal),
+  })
+
+  const [membersVisible, setMembersVisible] = useState(false)
```
jakehobbs: isMembersVisible

frontend-v2/src/app/(authed)/circles/delete-circle-dialog.tsx
```diff
@@ -0,0 +1,79 @@
...
+        </DialogHeader>
+        <p className="text-sm">
+          Are you sure you want to delete <strong>{circle.name}</strong>?
+        </p>
+        <div className="flex items-start gap-2 rounded-md border border-amber-300 bg-amber-50 p-3 text-sm text-amber-900">
+          <AlertTriangle className="mt-0.5 h-4 w-4 shrink-0" />
+          <p>
+            Before deleting a circle, be sure to remove all members of that
+            circle.
+          </p>
```
jakehobbs: we should really just enforce this client-side

frontend-v2/src/app/(authed)/circles/circle-form-dialog.tsx
jakehobbs: [Claude] Done in a8f0a734 — the members array is now built in one literal by spreading the optional host entry and the filtered/mapped members list (mirroring #464's construction in 3b340e8f). Ordering (host first) and host-dedup are unchanged, so the payload is identical.

frontend-v2/src/app/(authed)/circles/circle-table.tsx
jakehobbs: [Claude] Done in a8f0a734 — the columns are now one array literal ending in a conditional spread that picks the mode-specific column (Last Event for interest; Members or Total Members for geo), with the mode-specific columns hoisted to named consts inside the useMemo. No more push.

frontend-v2/src/app/(authed)/circles/circles-page.tsx
jakehobbs: [Claude] Done in a8f0a734 — renamed to isMembersVisible/setIsMembersVisible, and the CircleTable prop is renamed to isMembersVisible to match.

frontend-v2/src/app/(authed)/circles/delete-circle-dialog.tsx
jakehobbs: [Claude] Done in a8f0a734 — verified the backend behaves like working groups: model.DeleteCircleGroup rejects circles that still have circle_members rows ("Cannot delete circle because it has members associated with it", server/src/model/circles.go:289). Mirrored #464's fix: the Delete button is disabled while the circle has members, and the warning switches to explain that they must be removed before deletion.

alexsapps (approved):

PR #468, opened 2026-07-19
Add shared activist tag-input component

jakehobbs:
## Summary

- Two of the parallel Vue→React port PRs independently built near-duplicate chip/autocomplete multi-select components for picking activist names:
  - `jake/react-working-groups` (#464): `person-multi-select.tsx` (point person, members, non-members)
  - `jake/react-circles` (#466): `activist-tag-input.tsx` (host, members)
- This PR extracts a single shared `TagInput` component at `frontend-v2/src/components/tag-input.tsx` covering the union of both usages (single-select via `max=1`, unbounded multi-select, disabled state, optional built-in label vs. externally-supplied label).
- It's a **base PR**: the component is unused here (no callers yet) so CI/build has nothing to exercise it against. #464 and #466 are expected to rebase/adopt this component on top of this branch and delete their local duplicate files as follow-up commits, per the parent task that spawned all three PRs.

## Notes for adopters

- Filtering is case-insensitive **"starts with"**, matching legacy Vue `getFilteredActivists`/`getFilteredOrganizers` in `WorkingGroupList.vue` and `CirclesList.vue`. The Working Groups PR's `person-multi-select.tsx` had drifted to substring (`includes`) matching — adopting `TagInput` restores parity with the legacy app and circles' existing behavior. This is an intentional, documented behavior change for that one call site.
- Prop names differ slightly from both originals (unified on `max` instead of `max`/`maxItems`, added `maxSuggestions`, `label` is optional and only renders internally when passed).

## Test plan

- [x] `pnpm install`
- [x] `pnpm exec tsc --noEmit` (only pre-existing `baseUrl` deprecation warning)
- [x] `pnpm lint` (clean)
- [x] `pnpm build` (clean)
- [ ] #464 and #466 adopt `TagInput` and confirm no visual/behavioral regressions on Working Groups and Circles forms

### DIFF
--- a/frontend-v2/src/components/tag-input.tsx
+++ b/frontend-v2/src/components/tag-input.tsx
@@ -1,0 +1,209 @@
+'use client'
+
+import { useId, useMemo, useState } from 'react'
+import { X } from 'lucide-react'
+import { Label } from '@/components/ui/label'
+import { Popover, PopoverAnchor, PopoverContent } from '@/components/ui/popover'
+import { cn } from '@/lib/utils'
+
+export interface TagInputProps {
+  /**
+   * Currently selected values, rendered as removable chips. Order is
+   * preserved (new selections are appended).
+   */
+  value: string[]
+  /** Called with the full next array whenever a chip is added or removed. */
+  onChange: (next: string[]) => void
+  /**
+   * The full universe of selectable values (e.g. all activist names). The
+   * caller owns fetching/loading this list — this component does not fetch
+   * data itself, it only filters/renders what it's given.
+   *
+   * Values already present in `value` are excluded from suggestions.
+   */
+  options: string[]
+  /**
+   * Optional label rendered above the control via the shared `Label`
+   * primitive, wired up with `htmlFor`/`id`. Omit this if the caller wants
+   * to render its own external `<Label htmlFor={id}>` next to the control
+   * (pass `id` in that case so the label's `htmlFor` still resolves).
+   */
+  label?: string
+  /**
+   * `id` applied to the text input. Auto-generated if omitted. Only needed
+   * explicitly when an external label (not the built-in `label` prop) needs
+   * to reference this input via `htmlFor`.
+   */
+  id?: string
+  /** Placeholder shown in the text input while no chips are selected. */
+  placeholder?: string
+  /**
+   * Maximum number of selected values. Once reached, the text input is
+   * hidden (existing chips remain visible and, unless `disabled`,
+   * removable). Omit for unbounded selection. Pass `1` for single-select
+   * (point person, host, etc.) usage.
+   */
+  max?: number
+  /**
+   * Maximum number of matching suggestions shown in the dropdown at once.
+   * @default 20
+   */
+  maxSuggestions?: number
+  /** Disables the control: hides the text input and chip-remove buttons. */
+  disabled?: boolean
+}
+
+/**
+ * Chip/tag multi-select with type-ahead filtering over a fixed list of
+ * `options`, restricted to those options (there is no free-text/allow-new
+ * mode — this matches the legacy Vue `b-taginput` usages this component
+ * replaces, which all set `allow-new: false`).
+ *
+ * Filtering is case-insensitive "starts with", matching the legacy Vue
+ * `getFilteredActivists`/`getFilteredOrganizers` behavior (`WorkingGroupList.vue`,
+ * `CirclesList.vue`) that both the Working Groups and Circles pages ported from.
+ *
+ * Keyboard behavior:
+ * - `Enter` selects the first suggestion (no-op if there are none).
+ * - `Backspace` on an empty input removes the last chip.
+ *
+ * This is a presentational component: callers own data fetching (typically
+ * via a `useQuery` call that loads a plain name list once) and pass the
+ * result in as `options`.
+ */
+export function TagInput({
+  value,
+  onChange,
+  options,
+  label,
+  id,
+  placeholder = 'Search by name...',
+  max,
+  maxSuggestions = 20,
+  disabled = false,
+}: TagInputProps) {
+  const [text, setText] = useState('')
+  const [open, setOpen] = useState(false)
+  const generatedId = useId()
+  const inputId = id ?? generatedId
+
+  const atMax = max !== undefined && value.length >= max
+  const showInput = !disabled && !atMax
+
+  const suggestions = useMemo(() => {
+    const query = text.trim().toLowerCase()
+    if (!query) return []
+    return options
+      .filter(
+        (name) => !value.includes(name) && name.toLowerCase().startsWith(query),
+      )
+      .slice(0, maxSuggestions)
+  }, [options, text, value, maxSuggestions])
+
+  const addValue = (name: string) => {
+    if (!name.trim() || value.includes(name) || atMax) return
+    onChange([...value, name])
+    setText('')
+    setOpen(false)
+  }
+
+  const removeValue = (name: string) => {
+    onChange(value.filter((v) => v !== name))
+  }
+
+  const control = (
+    <div
+      className={cn(
+        'flex min-h-9 flex-wrap items-center gap-1.5 rounded-md border border-input bg-transparent px-2 py-1 text-sm',
+        'focus-within:border-primary focus-within:ring-1 focus-within:ring-ring',
+        disabled && 'cursor-not-allowed opacity-50',
+      )}
+    >
+      {value.map((name) => (
+        <span
+          key={name}
+          className="inline-flex items-center gap-1 rounded-full bg-secondary px-2 py-0.5 text-xs text-secondary-foreground"
+        >
+          {name}
+          {!disabled && (
+            <button
+              type="button"
+              onClick={() => removeValue(name)}
+              aria-label={`Remove ${name}`}
+              className="rounded-full hover:bg-secondary-foreground/10"
+            >
+              <X className="h-3 w-3" />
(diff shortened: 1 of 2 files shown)


frontend-v2/src/components/tag-input.tsx
```diff
@@ -0,0 +1,209 @@
...
+              id={inputId}
+              className="min-w-[8rem] flex-1 border-0 bg-transparent p-1 text-sm outline-none placeholder:text-muted-foreground"
+              value={text}
+              placeholder={value.length === 0 ? placeholder : undefined}
+              onChange={(e) => {
+                setText(e.target.value)
+                setOpen(true)
+              }}
+              onFocus={() => setOpen(true)}
+              onBlur={() => setOpen(false)}
```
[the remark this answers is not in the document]
jakehobbs: [Claude] Fixed in daced44e: the input now has `role="combobox"`, `aria-autocomplete`, `aria-expanded`, and `aria-controls`, linking it to the suggestions list. This mirrors the existing combobox pattern in `suggestion-input.tsx`.

frontend-v2/src/components/tag-input.tsx
```diff
@@ -0,0 +1,209 @@
...
+          >
+            <ul
+              role="listbox"
+              className="max-h-[240px] overflow-y-auto rounded-md border border-gray-200 bg-white shadow-lg"
+            >
+              {suggestions.map((s) => (
+                <li
+                  key={s}
+                  role="option"
+                  aria-selected={false}
```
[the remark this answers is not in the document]
jakehobbs: [Claude] Fixed in daced44e (same fix as the combobox comment above): the listbox now has a stable id and each option has an id, completing the `aria-controls`/`aria-activedescendant` linkage.

frontend-v2/src/components/tag-input.tsx
```diff
@@ -0,0 +1,209 @@
...
+        </Popover>
+      )}
+    </div>
+  )
+
+  if (!label) return control
+
+  return (
+    <div className="space-y-1.5">
+      <Label htmlFor={inputId}>{label}</Label>
```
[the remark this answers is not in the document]
jakehobbs: [Claude] Fixed in daced44e: the `<Label>`'s `htmlFor` is now only set while the input is actually rendered, so there's no dangling reference when `max` is reached or the field is disabled.

frontend-v2/src/components/tag-input.tsx
```diff
@@ -0,0 +1,209 @@
...
+        (name) => !value.includes(name) && name.toLowerCase().startsWith(query),
+      )
+      .slice(0, maxSuggestions)
+  }, [options, text, value, maxSuggestions])
+
+  const addValue = (name: string) => {
+    if (!name.trim() || value.includes(name) || atMax) return
+    onChange([...value, name])
+    setText('')
+    setOpen(false)
```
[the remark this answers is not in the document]
jakehobbs: [Claude] Acknowledged — keyboard-interaction test coverage is a good candidate to add, especially alongside the arrow-key navigation work tracked in issue #474. Not added in this PR to keep it reviewable.

frontend-v2/src/components/tag-input.tsx
```diff
@@ -0,0 +1,180 @@
...
+  /** All selectable values — callers own data fetching; already-selected values are excluded from suggestions. */
+  options: string[]
+  /** Optional label rendered above the control, wired to the input via `htmlFor`. */
+  label?: string
+  /** Input id for an external label (auto-generated if omitted); prefer `label`, since the input unmounts while hidden. */
+  id?: string
+  /** Placeholder shown in the text input while no chips are selected. */
+  placeholder?: string
+  /** Max selections — the input hides once reached; pass 1 for single-select, omit for unbounded. */
+  max?: number
```
jakehobbs: should we just have an option for single-select instead of max options if the only used values are binary (1 or no limit)?

frontend-v2/src/components/tag-input.tsx
```diff
@@ -0,0 +1,180 @@
...
+import { Popover, PopoverAnchor, PopoverContent } from '@/components/ui/popover'
+import { cn } from '@/lib/utils'
+
+export interface TagInputProps {
+  /** Selected values, rendered as removable chips (new selections are appended). */
+  value: string[]
+  /** Called with the full next array whenever a chip is added or removed. */
+  onChange: (next: string[]) => void
+  /** All selectable values — callers own data fetching; already-selected values are excluded from suggestions. */
+  options: string[]
```
jakehobbs, line 15: hmmm is this really just an array? is it at least pre-sorted? wonder if it worth be a worthwhile perf win to add context for these options that all tag inputs could share, so if that is an overoptimization. my frontend skills are rusty.

frontend-v2/src/components/tag-input.tsx
```diff
@@ -0,0 +1,180 @@
...
+  const showInput = !disabled && !atMax
+  // The input unmounts while hidden, so the label must not reference it then.
+  const labelFor = showInput ? inputId : undefined
+
+  const suggestions = useMemo(() => {
+    const query = text.trim().toLowerCase()
+    if (!query) return []
+    return options
+      .filter(
+        (name) => !value.includes(name) && name.toLowerCase().startsWith(query),
```
jakehobbs: this looks like terrible perf

frontend-v2/src/components/tag-input.tsx
```diff
@@ -0,0 +1,180 @@
...
+  options,
+  label,
+  id,
+  placeholder = 'Search by name...',
+  max,
+  maxSuggestions = 20,
+  disabled = false,
+}: TagInputProps) {
+  const [text, setText] = useState('')
+  const [open, setOpen] = useState(false)
```
jakehobbs: would prefer `isOpen` over `open`, unless this pattern is already established repo-wide

frontend-v2/src/components/tag-input.tsx
```diff
@@ -0,0 +1,180 @@
...
+  /** Selected values, rendered as removable chips (new selections are appended). */
+  value: string[]
+  /** Called with the full next array whenever a chip is added or removed. */
+  onChange: (next: string[]) => void
+  /** All selectable values — callers own data fetching; already-selected values are excluded from suggestions. */
+  options: string[]
+  /** Optional label rendered above the control, wired to the input via `htmlFor`. */
+  label?: string
+  /** Input id for an external label (auto-generated if omitted); prefer `label`, since the input unmounts while hidden. */
+  id?: string
```
jakehobbs: pretty confused what these ids are actually used for. if safe to remove, remove.

frontend-v2/src/components/tag-input.tsx
jakehobbs: [Claude] Good call — checked every call site on both adopting branches (`jake/react-working-groups`, `jake/react-circles`): all 5 usages are either `max={1}` (point person, host) or unlimited (members, non-members). Replaced `max?: number` with `single?: boolean` in 546908a3. Adopters change `max={1}` to `single`.

frontend-v2/src/components/tag-input.tsx
jakehobbs: [Claude] Assessed and left as-is — I think context would be overoptimization here. The option lists are name-only arrays (a few thousand entries at most — the full activist dataset with metadata is ~3MB per activist-storage.ts, and these name lists are a small fraction of that). Both pages already share the fetch across components via the react-query cache key, so context would only save prop-drilling one array one level deep. Filtering is a single O(n) prefix scan per keystroke, capped at 20 results — well under a millisecond at this n. Pre-sorting would actually be wrong: `GetAutocompleteNames` deliberately orders by most recent event attendance (`ORDER BY MAX(e.date) DESC`) so recently active people rank first in suggestions, same as the legacy Vue pages that use these endpoints.

frontend-v2/src/components/tag-input.tsx
jakehobbs: [Claude] Fixed in 546908a3. Before: the filter ran `value.includes(name)` per option — O(options x selected) per keystroke, plus it re-ran on every re-render. After: suggestions are computed in a `useMemo` keyed on (options, text, selected set, cap), with a memoized `Set` for the selected-value check — O(options) per keystroke and nothing on unrelated re-renders. At the actual data size (a few thousand short strings) this is comfortably sub-millisecond.

frontend-v2/src/components/tag-input.tsx
jakehobbs: [Claude] Renamed to `isOpen` in 546908a3. Grep confirms `isOpen` is the established convention for internal state: 3 existing uses (`suggestion-input.tsx`, `column-selector.tsx`, `debug-menu.tsx`) plus pervasive `is*` boolean state names, vs a single `const [open,` (`filters/filter-chip.tsx`). Radix component props (`<Popover open=...>`) are untouched — that name is the library API.

frontend-v2/src/components/tag-input.tsx
jakehobbs: [Claude] Cleaned up in 546908a3. Removed: the `id` prop (neither adopting branch passes it now that the built-in `label` prop handles wiring) and the per-option `id`s, which nothing referenced — they only matter for `aria-activedescendant` keyboard highlighting, which this component does not implement. What remains: one `useId()`-generated input id that the label points at via `htmlFor`, and a derived listbox id linking the input to the dropdown via `aria-controls` so screen readers associate the combobox with its popup.

alexsapps (approved):

PR #476, opened 2026-07-19
Port chapters page to React

jakehobbs:
## Summary

Ports the legacy chapters admin page (`frontend/ChapterList.vue`, `/list_chapters`, ~2250 lines) to the Next.js app at `/v2/chapters`. This is the largest remaining Vue page, managing DxE chapters worldwide (create/edit/delete/list, organizer sub-lists, mentor/name filtering, FB sync status).

No backend changes — reuses the existing `chapter/list`, `chapter/save`, and `chapter/delete` Go endpoints exactly as the Vue page does.

## Edit-UX decision: separate pages vs. drawer

Chose **separate pages** (`/v2/chapters`, `/v2/chapters/new`, `/v2/chapters/[id]`), matching the `users` precedent, over a right-side drawer.

Field count: 12 top-level fields (Name, Mentor, Notes, Region, Country, Lat, Lng, Facebook/Twitter/Instagram URLs, public Email, Flag derived from Country) plus a repeating Organizers sub-list (7 fields each: Name, Email, Phone, Facebook, Instagram, Twitter, Website).

Pros of separate pages (chosen):
- The Vue page itself disables adding organizers until the chapter is saved ("Please save the new chapter before adding organizers.") — a full page gives that follow-up step room to breathe, and after creating a chapter we navigate straight to its edit page so organizers can be added immediately.
- Deep-linkable URLs per chapter (useful since chapters are referenced from Users, International Organizers, and activist records).
- Matches the established `users` list + `/new` + `/[id]` pattern exactly, minimizing new conventions.
- More horizontal room for the organizer sub-list than a narrow drawer would allow.

Cons (accepted):
- An extra navigation hop compared to editing inline in the list (mitigated by TanStack Query cache reuse between list/edit — no extra fetch when going back).
- Loses the Vue modal's "quick edit without leaving the list" feel.

## Endpoints used

- `GET chapter/list` (existing, admin+intl_coordinator gated) — list + single-chapter lookup (there's no single-chapter GET endpoint, so the edit page fetches the full list and finds the chapter by ID, same as the Vue page keeping all chapters in memory).
- `POST chapter/save` — create (no `ChapterID`) or update (with `ChapterID`), same branch-by-presence logic as the Go handler already implements.
- `POST chapter/delete` — unchanged.

Added typed `ApiClient` methods (`getChapterAdminList`, `saveChapterAdmin`, `deleteChapterAdmin`) and `API_PATH.CHAPTER_SAVE` / `API_PATH.CHAPTER_DELETE` in `frontend-v2/src/lib/api.ts`, alongside a new `ChapterAdminSchema` (zod, `.passthrough()`) distinct from the existing minimal `Chapter`/`ChapterOrganizerSchema` used by the chapter-picker dropdown and `/v2/intl/organizers`.

## Feature parity / round-tripped fields

The Go `ChapterWithToken` struct has no JSON tags, so `chapter/save` overwrites every column in its `UPDATE` statement with whatever the client sends — including `email_token` and `last_checkin_email_sent`, two fields the Vue page never displays but silently preserves via `{...chapter}` object-spread before posting. A naive React port that only sent the displayed/edited fields would wipe these columns on every save. Fixed by giving `ChapterAdminSchema` `.passthrough()` and spreading the originally-fetched chapter into the save payload before applying form edits — the same effect as Vue's spread, without needing to model `sql.NullTime`'s JSON shape.

All Vue fields are ported: Name (locked after creation, matching `:disabled="currentChapter.ChapterID"`), Flag (still derived from Country selection, never directly editable), Mentor, Notes, Region, Country, Lat/Lng, Facebook/Twitter/Instagram URLs, public Email, Last Contact / Last Action (date picker + "Today"/"Clear" quick actions), and the Organizers sub-list (add/remove rows, all 7 fields). The 258-country dropdown (with flags) was extracted verbatim from the Vue file's `allCountries` array into `countries.ts`.

List page ports: New chapter button, Mentor filter, name filter, Show FB columns toggle, Total/Active chapter counts, Export CSV link (unchanged legacy `/csv/international_organizers` route), and per-row Edit/Email-compose/Delete actions. The Last Action color-coding (quadrimester-based green/red/black) and FB sync status coloring were ported algorithmically into `chapter-utils.ts`.

## Intentional differences

- Last Contact/Last Action edits happen inline in the full edit form rather than via separate one-off modals that auto-save on click.
- The organizer detail fields (Facebook/Instagram/Twitter/Website) are always visible in a grid per organizer instead of behind an expand/collapse row.
- After creating a new chapter, the user is redirected to its edit page (not back to the list) so organizers can be added right away.
- Lat/Lng of exactly `0` is accepted as valid (the Vue page's `!lat` check incorrectly rejected the equator).

- Admin > Chapters now points to `/v2/chapters` (`ChaptersList_beta`).
- Appended the old `/list_chapters` link to the end of the Legacy menu (same roles).

## Test plan

- [x] `pnpm exec tsc --noEmit` — passes (only the known/ignorable pre-existing `baseUrl` deprecation warning)
- [x] `pnpm lint` — 0 errors, 0 warnings
- [x] `pnpm build` — succeeds; `/chapters`, `/chapters/new`, `/chapters/[id]` all registered
- [x] `pnpm test` (vitest) — fails identically on `origin/main` with zero changes applied (pre-existing `ERR_REQUIRE_ESM` in a transitive jsdom dependency, unrelated to this change)
- [ ] Manual click-through against a dev DB (not run in this environment)


frontend-v2/src/app/(authed)/chapters/[id]/page.tsx
```diff
@@ -0,0 +1,46 @@
...
+export default async function EditChapterPage({
+  params,
+}: {
+  params: Promise<{ id: string }>
+}) {
+  const { id } = await params
+  const chapterId = parseInt(id)
+  if (Number.isNaN(chapterId)) {
+    notFound()
+  }
```
Comment (jakehobbs) [inline frontend-v2/src/app/(authed)/chapters/[id]/page.tsx:23]: should we use nuqs?

frontend-v2/src/app/(authed)/chapters/new/page.tsx
```diff
@@ -0,0 +1,19 @@
...
+import { getCookies } from '@/lib/auth'
+import { redirectForHttpError } from '@/lib/server-auth'
+import { ChapterForm } from '../chapter-form'
+
+export default async function NewChapterPage() {
+  const apiClient = new ApiClient(await getCookies())
+
+  // Not used by the form itself; this call exists purely to gate the page
+  // behind the same admin/intl_coordinator check as the rest of /chapters.
+  await redirectForHttpError(() => apiClient.getChapterAdminList())
```
jakehobbs: is this hacky? what did vue page do?

frontend-v2/src/app/(authed)/chapters/chapter-form.tsx
```diff
@@ -0,0 +1,755 @@
...
+  lng: z.number().optional(),
+  fbUrl: z.string().trim().max(100),
+  twitterUrl: z.string().trim().max(100),
+  instaUrl: z.string().trim().max(100),
+  email: z
+    .string()
+    .trim()
+    .max(100)
+    .refine(
+      (v) => v === '' || z.string().email().safeParse(v).success,
```
jakehobbs: doesn't zod have a `.email()` included?

frontend-v2/src/app/(authed)/chapters/chapter-form.tsx
```diff
@@ -0,0 +1,755 @@
...
+        toast.error('Please fix the highlighted errors.')
+        return
+      }
+
+      const flag =
+        COUNTRIES.find((c) => c.code === parsed.data.country)?.flag ?? ''
+
+      // Spread the original chapter first so fields the form never displays
+      // (e.g. EmailToken) round-trip unchanged instead of being wiped on save.
+      const payload: Partial<ChapterAdmin> = {
```
jakehobbs: should we extract a function for constructing payload?

frontend-v2/src/app/(authed)/chapters/chapter-form.tsx
```diff
@@ -0,0 +1,755 @@
...
+    email: chapter?.Email ?? '',
+    lastContact: chapter?.LastContact
+      ? parseDateYmd(chapter.LastContact)
+      : null,
+    lastAction: chapter?.LastAction ? parseDateYmd(chapter.LastAction) : null,
+    organizers: chapter?.Organizers ?? [],
+  }
+}
+
+function ChapterFormInner({
```
jakehobbs, line 18: this component is massive. pls consider splitting it up.

frontend-v2/src/app/(authed)/chapters/chapter-utils.ts
```diff
@@ -0,0 +1,131 @@
...
+
+function addMonths(date: Date, months: number): Date {
+  return new Date(date.getFullYear(), date.getMonth() + months, date.getDate())
+}
+
+function addDays(date: Date, days: number): Date {
+  const result = new Date(date)
+  result.setDate(result.getDate() + days)
+  return result
+}
```
jakehobbs: if we have date-fns installed, pls use it whenever possible

frontend-v2/src/app/(authed)/chapters/chapter-utils.ts
```diff
@@ -0,0 +1,131 @@
...
+  const date = parseDateYmd(value)
+  if (date == null) return false
+  const threeMonthsAgo = new Date()
+  threeMonthsAgo.setMonth(threeMonthsAgo.getMonth() - 3)
+  return date > threeMonthsAgo
+}
+
+// Quadrimesters: Feb-May, Jun-Sep, Oct-Jan. Returns the first day of the
+// quadrimester containing today.
+function currentQuadrimesterStart(now: Date): Date {
```
jakehobbs, line 45: what's this? i don't remember quadrimesters in vue, but we need behavior to match vue. was this supposed to be quarter?

frontend-v2/src/app/(authed)/chapters/chapters-page.tsx
```diff
@@ -0,0 +1,195 @@
...
+          <Loader2 className="h-4 w-4 animate-spin" />
+          Loading chapters...
+        </div>
+      ) : (
+        <ChapterTable
+          chapters={filteredChapters}
+          showFacebookColumns={showFacebookColumns}
+          onDelete={handleDelete}
+        />
+      )}
```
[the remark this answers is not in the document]
jakehobbs: [Claude] Fixed in 89497933 — the list page now renders an explicit error notice (`error.message`, falling back to a generic message) between the loading and success states, mirroring `users-page.tsx`/`organizers-page.tsx`, instead of falling through to an empty table.

frontend-v2/src/app/(authed)/chapters/chapter-utils.ts
```diff
@@ -0,0 +1,131 @@
...
+  Email: string
+  Organizers: { Email: string }[]
+}): string | null {
+  const emails = [
+    chapter.Email,
+    ...chapter.Organizers.map((o) => o.Email),
+  ].filter(Boolean)
+  if (emails.length === 0) return null
+  return `https://mail.google.com/mail/?view=cm&fs=1&su=${chapter.Name}&to=${emails.join(',')}`
+}
```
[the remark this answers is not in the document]
jakehobbs: [Claude] Fixed in 89497933 — the compose link is now built with `URLSearchParams` (`view`, `fs`, `su`, `to`), so the chapter name and joined emails are properly encoded.

frontend-v2/src/app/(authed)/chapters/chapter-table.tsx
```diff
@@ -0,0 +1,239 @@
...
+  STATUS_COLOR_CLASSES,
+} from './chapter-utils'
+
+function composeChapterEmail(chapter: ChapterAdmin) {
+  const link = buildChapterEmailLink(chapter)
+  if (link == null) {
+    toast.error(`There are no email addresses listed for ${chapter.Name}!`)
+    return
+  }
+  window.open(link)
```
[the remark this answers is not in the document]
jakehobbs: [Claude] Fixed in 89497933 — now `window.open(link, '_blank', 'noopener,noreferrer')`.

frontend-v2/src/app/(authed)/chapters/countries.ts
```diff
@@ -0,0 +1,268 @@
...
+  { name: 'Singapore', code: 'SG', flag: '🇸🇬' },
+  { name: 'St. Helena', code: 'SH', flag: '🇸🇭' },
+  { name: 'Slovenia', code: 'SI', flag: '🇸🇮' },
+  { name: 'Svalbard & Jan Mayen', code: 'SJ', flag: '🇸🇯' },
+  { name: 'Slovakia', code: 'SK', flag: '🇸🇰' },
+  { name: 'Sierra Leone', code: 'SL', flag: '🇸🇱' },
+  { name: 'San Marino', code: 'SM', flag: '🇸🇲' },
+  { name: 'Senegal', code: 'SN', flag: '🇸🇳' },
+  { name: 'Somalia', code: 'SO', flag: '🇸🇴' },
+  { name: 'SuriName', code: 'SR', flag: '🇸🇷' },
```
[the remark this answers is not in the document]
jakehobbs: [Claude] Fixed in 89497933. Verified this is safe first: the typo was carried over verbatim from the Vue source (frontend/ChapterList.vue line 1978, `Name: 'SuriName'`), but the persisted value is the country *code* — the Vue select binds `:value="country.Code"` (line 265) — so `fb_pages.country` stores `SR`, never the display name. Fixing the label cannot orphan existing rows.

frontend-v2/src/app/(authed)/chapters/[id]/page.tsx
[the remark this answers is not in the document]
Comment (jakehobbs) [reply frontend-v2/src/app/(authed)/chapters/[id]/page.tsx]: [Claude] Fixed in 89497933 — the id segment is now validated with a strict `/^\d+$/` test before `Number(id)`; anything else 404s via `notFound()`.

Comment (jakehobbs) [reply frontend-v2/src/app/(authed)/chapters/[id]/page.tsx:23]: [Claude] I do not think nuqs applies here: it manages URL *query-string* state (filters, search params) synced to client components, whereas this is a dynamic route segment (`params.id`) resolved by the App Router on the server. There is no query state on this page to manage, so I kept the standard `params` handling (now with strict integer validation per the sibling comment).

frontend-v2/src/app/(authed)/chapters/new/page.tsx
jakehobbs: [Claude] Agreed it was hacky — replaced in 89497933. The Vue page was gated server-side by `authIntlCoordinatorAccessMiddleware` on the `/list_chapters` HTML route (server/src/main.go:306); the API calls are separately gated by `apiIntlCoordinatorAuthMiddleware`. The `/new` page now uses the established session gate instead of a throwaway fetch: `getCachedSession()` + `forbidden()` unless the user has `admin` or `intl_coordinator` — mirroring `intlCoordinatorRoles` in server/src/model/adb_auth.go:12 — same as `config/page.tsx` does for admin. The list and `[id]` pages keep their prefetch + `redirectForHttpError` since they genuinely need the data anyway.

frontend-v2/src/app/(authed)/chapters/chapter-form.tsx
jakehobbs: [Claude] Yes — replaced the manual refine in 89497933 with `z.string().trim().max(100).pipe(z.email('Enter a valid email').or(z.literal('')))` (zod v4 top-level `z.email()`). Same semantics: trims, accepts empty (the field is optional), rejects invalid addresses.

frontend-v2/src/app/(authed)/chapters/chapter-form.tsx
jakehobbs: [Claude] Done in 89497933 — extracted `buildChapterPayload(chapter, values)` into `chapter-form-schema.ts` next to the schemas, which also gives the round-trip comment (spread original chapter so undisplayed fields like EmailToken are not wiped) a natural home.

frontend-v2/src/app/(authed)/chapters/chapter-form.tsx
jakehobbs: [Claude] Done in 89497933 — split following the events form layout: `chapter-form-schema.ts` (zod schemas, `toInitialValues`, `buildChapterPayload`), `useChapterForm.ts` (mutation + form hook, exports `ChapterFormApi` like `useEventForm`), and `chapter-form-sections.tsx` (Basic Info / Location / Social Links / Dates / Organizers section components, with the repeated coordinate/social/date/organizer fields built declaratively from config arrays). `chapter-form.tsx` is now ~100 lines of header + submit + section composition. Behavior unchanged.

frontend-v2/src/app/(authed)/chapters/chapter-utils.ts
jakehobbs: [Claude] Done in 89497933 — `chapter-utils.ts` now uses date-fns (`parse`/`format` for the Y-M-D round-trip, `addMonths`/`subMonths`/`addDays`/`subDays`/`addHours`/`subHours` for the threshold math, `isAfter`/`isBefore`/`isValid` for comparisons, `differenceInCalendarDays` for the tooltip) instead of the hand-rolled helpers.

frontend-v2/src/app/(authed)/chapters/chapter-utils.ts
jakehobbs: [Claude] Quadrimesters genuinely come from the Vue page, not from me — see frontend/ChapterList.vue lines 853-883: the comment "Quadrimesters: Feb–May, Jun–Sep, Oct–Jan" (853-854), `currentQuadrimesterStart()` (855-863), and `colorLastAction()` (864-883) with the same prev-quad / +1mo+2wk red threshold / quadEnd-1wk black threshold logic this file ports 1:1. It was added to Vue recently, which may be why it is unfamiliar: commit 359a1b9c "feat(chapters): last action coloring" (May 16, 2026) and follow-up efbc75c1 "fix(chapters): last action coloring" (May 18, 2026), both by Alexander Taylor. So it is intentionally 4-month blocks, not quarters, and React matches Vue exactly (verified again after the date-fns rewrite in 89497933).

frontend-v2/src/app/(authed)/chapters/countries.ts
```diff
@@ -0,0 +1,268 @@
+export interface Country {
+  name: string
+  code: string
+  flag: string
+}
+
+// Extracted from frontend/ChapterList.vue's allCountries list (display names
+// only; the persisted value is `code`, so fixing Vue's "SuriName" typo is safe).
```
alexsapps, line 8:
```suggestion
```

frontend-v2/src/app/(authed)/chapters/chapter-form-schema.ts
```diff
@@ -0,0 +1,105 @@
...
+    .trim()
+    .max(100)
+    .pipe(z.email('Enter a valid email').or(z.literal(''))),
+  lastContact: z.date().nullable(),
+  lastAction: z.date().nullable(),
+  organizers: z.array(organizerFormSchema),
+})
+
+// Stricter checks that only apply at submit time, mirroring the legacy Vue
+// page's confirmEditChapterModal validation.
```
alexsapps, line 40:
```suggestion
// Stricter checks that only apply at submit time.
```

frontend-v2/src/app/(authed)/chapters/chapter-table.tsx
alexsapps, line 1: the columns here stretch to fill the screen and leave lots of blank space between columns. perhaps the columns should have max widths.

alexsapps (approved):

frontend-v2/src/app/(authed)/chapters/chapter-table.tsx
alexsapps: or maybe make it a non-full-screen page, but i forget maybe we're trying to make all the pages full screen so we can get rid of the background image for performance reasons

frontend-v2/src/app/(authed)/chapters/chapters-page.tsx
alexsapps, line 1: was it intentional to drop the cards view for mobile that Vue had? now on mobile it's a horizontally scrolling table

frontend-v2/src/app/(authed)/chapters/chapter-form-schema.ts
```diff
@@ -0,0 +1,105 @@
...
+    lastContact: chapter?.LastContact
+      ? parseDateYmd(chapter.LastContact)
+      : null,
+    lastAction: chapter?.LastAction ? parseDateYmd(chapter.LastAction) : null,
+    organizers: chapter?.Organizers ?? [],
+  }
+}
+
+// Spreads the original chapter first so fields the form never displays (e.g.
+// EmailToken) round-trip unchanged instead of being wiped on save.
```
alexsapps, line 82: i did a lil more looking into EmailToken and it looks like it's not used anymore. maybe we could at least update the comment to mention this / add a TODO.

Claude said:

---

/international_actions — it's dead
grep-ing all of server/src for "international_actions" finds nothing. There's no HandleFunc/Handle registration for that path in main.go, and no path-prefix mounting would catch it either — the only prefixes registered are for static assets (/static, /dist, /v2, /_next, etc.).

Git history explains it: the whole feature was built in commit f387daa8 (2021), then fully removed in commit e26251e2 ("remove international action form", 2023-06-25) — but that commit only cleaned up main.go, international_mailer/international_mailer.go, model/forms.go, and frontend/adb.ts. It left behind:

frontend/FormInternationalActions.vue (the file you're looking at)
server/templates/form_international_actions.html
the form_international_actions DB table
the email_token / EmailToken plumbing in chapters.go
So POSTing to /international_actions today just 404s. The Vue component is orphaned dead code.

For reference, before removal the route/handler looked like:


router.HandleFunc("/international_actions/{id:[0-9]+}/{token:[a-zA-Z0-9]+}", main.InternationalActionsFormHandler)
router.HandleFunc("/international_actions", main.InternationalActionsFormHandler)
The GET side loaded a chapter by ID and checked chapFromDB.EmailToken != token; the POST side decoded InternationalActionFormData and called model.SubmitInternationalActionForm.

EmailToken — its purpose
It's a per-chapter secret capability token, generated once at chapter creation (generateToken() at [main.go:67-74](vscode-webview://1qjq7s6paih490fgp7qjb9l49l5ukckghn22ooqo18oanfforr1v/server/src/main.go#L67-L74), assigned in ChapterSaveHandler at [main.go:866](vscode-webview://1qjq7s6paih490fgp7qjb9l49l5ukckghn22ooqo18oanfforr1v/server/src/main.go#L866) — this still runs today even though nothing consumes it anymore).

It was used to build no-login, unguessable URLs like https://adb.dxe.io/international_actions/{chapterID}/{emailToken}, emailed monthly to chapter organizers so they could submit a check-in update without an ADB account. The GET handler validated the token by equality check against the DB value; on match it rendered the check-in form — the exact form in FormInternationalActions.vue, which passes token straight through as part of its POST body ([FormInternationalActions.vue:110](vscode-webview://1qjq7s6paih490fgp7qjb9l49l5ukckghn22ooqo18oanfforr1v/frontend/FormInternationalActions.vue#L110)).

It pairs with LastCheckinEmailSent ([chapters.go:74](vscode-webview://1qjq7s6paih490fgp7qjb9l49l5ukckghn22ooqo18oanfforr1v/server/src/model/chapters.go#L74)) — that's why the two columns are still selected/updated together in GetAllChapters, GetAdminChapterById, UpdateChapter, and InsertChapter today. Both are leftovers of the same removed workflow; the mailer that sent those emails (international_mailer.go) was deleted in the same 2023 cleanup commit.

Bottom line: if you're trying to make this form work again, the backend route/handler needs to be re-added (git history has the old implementation to restore). If instead you're trying to finish the 2023 cleanup, the Vue component, the HTML template, the form_international_actions table, and the email_token/EmailToken field are all candidates for removal.
```suggestion
// Spreads the original chapter first so fields the form never displays (e.g.
// EmailToken) round-trip unchanged instead of being wiped on save.
// Note that EmailToken is no longer used and should be removed anyway.
```

frontend-v2/src/lib/api.ts
```diff
@@ -143,6 +145,58 @@ export function flattenChapterOrganizers(
   )
 }
 
+// Full `fb_pages` row for the chapters admin page. `.passthrough()` preserves
+// fields the UI never edits (e.g. EmailToken) so saves can't silently wipe them.
```
alexsapps, line 149: here too
```suggestion
// fields the UI never edits (e.g. EmailToken) so saves can't silently wipe them.
// Note that EmailToken is unused and should be removed anyway.
```

frontend-v2/src/lib/api.ts
```diff
@@ -143,6 +145,58 @@ export function flattenChapterOrganizers(
...
+    Mentor: z.string(),
+    Notes: z.string(),
+    LastContact: z.string(),
+    LastAction: z.string(),
+    Organizers: z
+      .array(ChapterOrganizerAdminSchema)
+      .nullable()
+      .transform((v) => v ?? []),
+  })
+  .passthrough()
```
alexsapps, line 185: (passthrough called here)

```suggestion
  // todo: remove when EmailToken is removed
  .passthrough()
```

frontend-v2/src/app/(authed)/chapters/chapters-page.tsx
jakehobbs: nope

frontend-v2/src/app/(authed)/chapters/countries.ts
jakehobbs: [Claude] Done in 5d805196 — trimmed to a one-liner: `// Extracted from frontend/ChapterList.vue's allCountries list.`

frontend-v2/src/app/(authed)/chapters/chapter-form-schema.ts
jakehobbs: [Claude] Applied in 5d805196.

frontend-v2/src/app/(authed)/chapters/chapter-form-schema.ts
jakehobbs: [Claude] Done in 5d805196:

```ts
// Spreads the original chapter first so fields the form never displays
// round-trip unchanged instead of being wiped on save.
// TODO: the only such field is EmailToken, which is unused and should be removed.
```

frontend-v2/src/lib/api.ts
jakehobbs: [Claude] Applied in 5d805196.

frontend-v2/src/lib/api.ts
jakehobbs: [Claude] Applied in 5d805196.

frontend-v2/src/app/(authed)/chapters/chapter-table.tsx
jakehobbs: [Claude] Fixed in 5d805196 — took your second option, and it turns out it was a porting mistake on my part rather than a styling choice.

Vue rendered this page in `.body-wrapper-extra-wide`: centered, `width: 95%`, `max-width: 1450px` (frontend/static/css/style.css:67-71) — *not* full-screen. I had ported it as `<ContentWrapper size="full">`, which is `lg:max-w-none lg:mx-0 lg:rounded-none shadow-none mb-0` — hence the columns spreading across the whole viewport.

Two independent signals that `full` was wrong here:

- `ContentWrapper`'s own docs say "Be sure to keep in sync with `isFullScreenPage` in site-background-controller.tsx", and that list is only `/activists` and `/intl/organizers` — `/chapters` was never added. So the page was rendering edge-to-edge *while still painting the site background*, which is the combination `full` exists to avoid.
- Vue's only other extra-wide pages are WorkingGroupList and FacebookEvents, neither of which is full-screen in v2 either.

So `page.tsx` and `loading.tsx` are now `<ContentWrapper size="2xl">` (`lg:max-w-screen-xl` → 1536px, the closest Tailwind step to Vue's 1450px), which bounds the table without needing per-column max widths. To your aside: we are not making all pages full-screen — `full` is reserved for the two genuinely edge-to-edge grids above, and dropping the background image is a side effect of those, not the goal.

I left per-column `max-w-*` off deliberately: with the wrapper bounded, capping columns just moves the slack rather than removing it, and truncating Name/Mentor would lose text Vue showed in full. Happy to add caps on top if it still reads sparse to you at 1536px.

frontend-v2/src/app/(authed)/chapters/chapters-page.tsx
jakehobbs: [Claude] Not intentional — restored in 5d805196. Buefy's `b-table` defaults `mobile-cards` to `true`, so Vue got the card layout for free and I lost it by hand-rolling the table.

`chapter-table.tsx` now renders the same two-branch layout the other v2 tables already use (`hidden md:block` table + `md:hidden space-y-3` cards) — see `users/user-table.tsx:171,217` and `events/event-list-table.tsx:205,264`. Both branches read from the same `table.getRowModel().rows`, so sorting and the mentor/name filters apply identically and the FB columns still toggle with `showFacebookColumns`.

Card layout: flag + name as the heading with the edit/email/delete icon buttons on the right, then a `<dl>` of Mentor / Last Contact / Last Action (keeping the `colorLastAction` pill and its tooltip) plus Last FB Event / FB Sync Status when FB columns are on. Empty state is "No chapters found." in both branches.

PR #508, opened 2026-09-04
feat: make reminder page list members for current month

mdzirbel:
Antonelle said seeing the current month would be more helpful for this

### DIFF
--- a/server/src/members/reminder.go
+++ b/server/src/members/reminder.go
@@ -22,7 +22,6 @@ func (s *server) reminder() {
 	}
 
 	year, month, _ := time.Now().Date()
-	month++ // next month
 	if month == time.December+1 {
 		year, month = year+1, time.January // carry to next year
 	}


server/src/members/reminder.go
```diff
@@ -22,7 +22,6 @@ func (s *server) reminder() {
 	}
 
 	year, month, _ := time.Now().Date()
-	month++ // next month
```
alexsapps, line 25: sometimes the chapter meeting is on the first day of the month, such as the most recent chapter meeting (tuesday september 1st, 2026) leaving very little time to get people to sign the voting agreement. maybe the email should include both the current and next month?

alexsapps (approved):

server/src/members/reminder.go
mdzirbel: I'll check with Antonelle! I'll make that a followup if so

